PR #168, opened 2017-12-12
Not needed dependency

VeraZab:


bpostlethwaite:
:dancer:

nicolaskruchten:
I'm pretty sure this dependency is still needed, at least as a dev dep: https://github.com/plotly/react-plotly.js-editor/blob/master/examples/simple/src/App.js#L3

VeraZab:
this is why it's in the package.json of examples/simple: https://github.com/plotly/react-plotly.js-editor/blob/master/examples/simple/package.json#L9

nicolaskruchten:
oooh right, I had forgotten that the examples have their own package.json etc, sorry!

PR #170, opened 2017-12-14
Create combined translation keys file

VeraZab:
@bpostlethwaite @alexcjohnson for your review


package.json
```diff
@@ -13,18 +13,26 @@
     "url": "https://github.com/plotly/react-plotly.js-editor/issues"
   },
   "scripts": {
+    "combine-translation-strings": "babel-node utils/combineTranslationKeys.js",
     "find-strings": "babel-node utils/findLocaleStrings.js",
```
bpostlethwaite: actually find-strings could be changed to `npm run make:translations`

bpostlethwaite: or better
`npm run make:locale` or `npm run make:localization`

package.json
```diff
@@ -13,18 +13,26 @@
     "url": "https://github.com/plotly/react-plotly.js-editor/issues"
   },
   "scripts": {
+    "combine-translation-strings": "babel-node utils/combineTranslationKeys.js",
```
bpostlethwaite: `npm run make:combined-locales` ?

Just so it fits in with all the other make stuff?

package.json
alexcjohnson: we're not making translations or localizations, we're making the list of strings from which someone would make the required translations

bpostlethwaite: or whatever. `make:something` ;)

bpostlethwaite:
Since this 'makes' another build artifact required to run in prepublish you could give it a `make:?` name.

:dancer: other than that!

package.json
bpostlethwaite: `make:translation-strings`

VeraZab:
haha this approval button's pretty strict, another 👀  @bpostlethwaite ?

utils/combinedLocales.js
```diff
@@ -3,7 +3,7 @@ import fs from 'fs';
 
 const pathToCombinedTranslationKeys = path.join(
   __dirname,
-  'combinedTranslationKeys.txt'
+  'combinedLocales.txt'
```
bpostlethwaite: Sorry to nitpick but as @alexcjohnson suggested perhaps we should continue calling these `translation-references` or `translation-strings` or something. Locales are the full objects with all the fancy dateFormats and such right?

VeraZab: plotly.js calls it's file translation-keys, I could call it translation-keys for the editor too, and combined-translation-keys for the combined file, works?

alexcjohnson: plotly.js has a rule of all-lowercase filenames, which seems to have fallen out of fashion in more recent js projects, but I won't vote against consistency across repos, I like your proposed names @VeraZab.

bpostlethwaite: sg!

VeraZab:
ok, how about those filenames? and also changed some scripts around

PR #173, opened 2017-12-16
async-data-example

bpostlethwaite:
@nicolaskruchten @VeraZab this example is closer to the use case of a particular client's application. I think it is an important example as many real-world cases involve fetching data asynchronously. 

Our current system using the hub and revisions is obviously cumbersome. I think this PR may make a good base to fix https://github.com/plotly/react-plotly.js-editor/issues/172.


examples/async-data/src/App.js
```diff
@@ -0,0 +1,186 @@
...
+    // The hub listens for events from plotly.js and alerts us to the mutation here.
+    // The Editor follows the same mutation pattern (for good or ill) and the Hub
+    // will pick up editor results in the same way.
+    this.hub = new Hub({
+      debug: true,
+      onUpdate: ({editorRevision, plotRevision}) =>
+        this.setState(prevState => ({
+          ...prevState,
+          editorRevision,
+          plotRevision,
```
bpostlethwaite: why does a user have to manage these things? Because we mutate data? I don't think that is a good enough excuse.

examples/async-data/src/App.js
```diff
@@ -0,0 +1,186 @@
...
+    backend.off('ChartingData', this.setChartingData);
+  }
+
+  setChartingDataOptions(columnNames) {
+    const dataSourceOptions = columnNames.map(name => ({
+      value: name,
+      label: name,
+    }));
+    this.setState(prevState => ({...prevState, dataSourceOptions}));
+    this.hub.editorSourcesUpdated();
```
bpostlethwaite: really? So a user needs to know what is and what is an editorSource and in each case make sure the hub is available in scope to call this function? really?

examples/async-data/src/App.js
```diff
@@ -0,0 +1,186 @@
...
+  setChartingData({columnName, data}) {
+    if (Array.isArray(data) && data.length) {
+      const {dataSources, ...otherState} = this.state;
+      dataSources[columnName] = data;
+      dereference(this.state.figure.data, dataSources);
+      this.setState({
+        dataSources,
+        ...otherState,
+      });
+      this.hub.figureUpdated();
```
bpostlethwaite: Oh and it's not just `this.hub.editorSourcesUpdated()` you need to keep track of, it's also somehow different from whatever `figureUpdated()` does. Nope and Nope.

examples/async-data/src/App.js
```diff
@@ -0,0 +1,186 @@
...
+    }
+  }
+
+  onEditorUpdate(event) {
+    const {type, payload} = event;
+    if (type === EDITOR_ACTIONS.UPDATE_TRACES) {
```
bpostlethwaite: And if we need to intercept the figure updating routine you gotta know how EDITOR_ACTIONS works and which events to tie into ... please :see_no_evil:

bpostlethwaite:
Okay a few scathing comments made. Extra scathing because I am solely responsible for this travesty. Who wants to take lead on figuring out and implementing a better system?

examples/async-data/src/App.js
```diff
@@ -0,0 +1,186 @@
...
+    // react-plotly.js plot component respectively. This is necessary because
+    // the plotly.js library will mutate the figure object with user interactions.
+    // The hub listens for events from plotly.js and alerts us to the mutation here.
+    // The Editor follows the same mutation pattern (for good or ill) and the Hub
+    // will pick up editor results in the same way.
+    this.hub = new Hub({
+      debug: true,
+      onUpdate: ({editorRevision, plotRevision}) =>
+        this.setState(prevState => ({
+          ...prevState,
```
nicolaskruchten: @bpostlethwaite why do we do this spread thing here? My understanding of `setState` was that it would merge anyway, no?

nicolaskruchten: see: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-are-merged

examples/async-data/src/App.js
nicolaskruchten: can this be done in the editor with `componentWillReceiveProps()`?

examples/async-data/src/App.js
nicolaskruchten: here we could invert control and have the hub fire its own sub-events that apps could hook into, so this would be implemented as `hub.onUpdateTrace = ...`

examples/async-data/src/App.js
bpostlethwaite: oh yes good call

examples/async-data/src/App.js
bpostlethwaite: can we eliminate the hub entirely? It still seems a bit ungainly to me. It's a stateful thing but distinct from typical application data handling.

examples/async-data/src/App.js
bpostlethwaite: it can if figure updates always generate new wrapping containers.

Issue #172, opened 2017-12-15
data management is currently a PITA. fix it.


bpostlethwaite:
I spent time integrating this library into a codebase and managing data and what updates when was a real PITA. We currently do not have a consistent or sane way to make this happen.

I am pretty unsure about the `hub` generally. In a larger app like Plotly's frontend where would the `hub` live exactly? I think we should try and :hocho: it.

Should we replace it with a collection of actions and reducers that people can use in their own apps data flow patterns? 

Should we remove `editorRevision` and `plotRevision` from the hub (or the whole hub) and also remove the `revision` props from `react-plotly.js` and `react-plotly.js-editor`? Instead we could expect shallow equals checks of props.

In some cases it may be necessary to mutate a giant data array. In that case we could demand that the `dataSources` object or the plotly.js `gd.data` object will be "rewrapped" in a new container therefore triggering the correct behavior in a shallow equals check. 

The reducers we supply could encode that behaviour.

nicolaskruchten:
OK so #176 improves the situation somewhat by 🔪 ing the Hub and grafting the pieces onto the Editor, but we retained the revision numbers. We could give things another turn and use shallow comparisons on `graphDiv` and `data` and `layout` to trigger updates, but I'll leave that for a second PR, as it will likely require changes to `react-plotly.js` as well.

bpostlethwaite:
Great I'm happy to close this issue for now.

PR #175, opened 2017-12-19
WIP Async with merged hub

nicolaskruchten:
This doesn't fully work just yet but it's the direction I'm thinking about for #172


bpostlethwaite:
:+1: for the general direction. Perhaps we can get a few tests on that hub/data update behaviour now that it is more stable?

PR #176, opened 2017-12-19
Merging Hub into Editor

### LINKED ISSUE #172: data management is currently a PITA. fix it.

nicolaskruchten:
Here's #175 retargeted onto #173 ... Meant to fix #172 :)


src/PlotlyEditor.js
```diff
@@ -35,23 +44,95 @@ class PlotlyEditor extends Component {
...
       locale: this.props.locale,
-      onUpdate: this.updateProp.bind(this),
+      onUpdate: this.handleUpdate.bind(this),
       plotSchema: this.plotSchema,
       plotly: this.props.plotly,
     };
   }
 
-  updateProp(event) {
+  handleUpdate({type, payload}) {
```
nicolaskruchten: We could go one step further here and split this handler into `handleUpdateTraces`, `handleDeleteAnnotation` etc... there's essentially no shared code here and no real need to have a standard `{type, payload}` pattern IMO.

src/PlotlyEditor.js
```diff
@@ -35,23 +44,95 @@ class PlotlyEditor extends Component {
...
+      default:
+        throw new Error('must specify an action type to handleEditorUpdate');
     }
   }
 
   render() {
     return (
       <div className={bem()}>
         {this.props.graphDiv &&
+          this.props.graphDiv._fullLayout &&
```
nicolaskruchten: this is a big change, conceptually, because it allows us to pass a partially-constructed figure into the editor and the plot, then the plot finishes constructing it.

examples/async-data/src/App.js
```diff
@@ -58,37 +54,22 @@ class App extends Component {
     super();
 
     // A basic starting plotly.js figure object. Instead of assigning
-    const figure = {
+    const graphDiv = {
       data: [{type: 'scatter'}],
       layout: {title: 'Room readings'},
     };
```
VeraZab: I'm not sure about this, when the plot component updates, it sets graphDiv to the actual dom element:
https://github.com/plotly/react-plotly.js/blob/master/src/factory.js#L167

it's that actual dom element that we're passing around and deriving all our values from, so making it seem like it is not the case is confusing to me

@bpostlethwaite is this not the case, am I missing something here?

VeraZab: Also, I don't think I understand how that state.figure used to work and update, what was making it update?

nicolaskruchten: Plotly.js would basically copy the `figure.data` and `figure.layout` references into `graphDiv` and pass it to the Editor via the Hub... Now it's just more explicit that the App has a handle on `graphDiv`: it partially-constructs it (i.e. gives it `data` and `layout`) and then Plotly.js adds all the other stuff it needs to  (basically by overwriting it every time it updates)

VeraZab: but to me it seems like the graphDiv that plotly.js passes is the actual dom div element.. when looking at the code

VeraZab: https://github.com/plotly/react-plotly.js/blob/master/src/factory.js#L189

nicolaskruchten: It is, but that dom div element has `data` and `layout` and `_fullLayout` etc attached to it as well...

VeraZab: yes I know :) the code here hides it a little, makes it seem like graphDiv is just an object, when that object is later changed to an actual dom node from which data, layout and everything else's derived. 

I'm ok with how it is used in instantiation here, but let's say I'd prefer we don't rename it like you're suggesting here: https://github.com/plotly/react-plotly.js-editor/issues/178

I think that editor users will most likely also be familiar with plotly.js and won't be surprised about our use of graphDiv and the fact that its a dom node that has all these properties

VeraZab: maybe we could add a note about it in here

nicolaskruchten: Let's continue the conversation in #178 :)

nicolaskruchten: Re the code here, we could also just have an `initialFigure = {data, layout}` type thing, with the initial value of `graphDiv` as `null`... I had it this way earlier and it worked as well, if we want to make it really explicit what's happening.

nicolaskruchten: what note would you add?

VeraZab: that graphDiv is later on replaced by a full dom node with all the other properties of plotly.js

VeraZab: > Re the code here, we could also just have an initialFigure = {data, layout} type thing, with the initial value of graphDiv as null
I think I prefer that

Then no note is really needed :)

nicolaskruchten: I added a comment per your suggestion. The reason I removed the `initialFigure` thing is that in `render()` it looked really weird to have `<Plot data={this.initialFigure.data} layout={this.initialFigure.layout} ... />` because it seemed like it would never get new data.

VeraZab:
yup, alright the comment works for me, so on my end, we're good on this one

nicolaskruchten:
OK, awesome :)

Any thoughts on breaking up `handleUpdate` or the rest of architectural changes?

VeraZab:
I don't mind `handleUpdate` staying as is for now

VeraZab:
maybe add react-plotly.js-editor in the package.json of the async example? ;)

nicolaskruchten:
added to package.json

nicolaskruchten:
@bpostlethwaite any additional comments or does this work for you, in terms of addressing your worst complaints about #173 ?

src/PlotlyEditor.js
bpostlethwaite: Is there a case where users will want to have their own update function? What if they want the updates to happen inside some reducer file, or apply updates and also do other things... If there are multiple functions it may be more difficult to replace?

bpostlethwaite:
Yep a single update function works for me. I am not sure how many users will need to override this behaviour yet or if that is a total non-issue. 

:dancer: !!!

PR #177, opened 2017-12-19
Standardize on widget field instead of unconnected widget

bpostlethwaite:
@VeraZab @nicolaskruchten 

Quick PR renaming and exposing unconnected components. Useful to my current project so probably generally useful.


VeraZab:
💃

bpostlethwaite:
I don't think I will actually need these components. Closing PR for now.

PR #179, opened 2017-12-21
Adding Redux example

VeraZab:


VeraZab:
ok @bpostlethwaite , the example works now, however it seems like `revisions` are necessary.

I didn't quite figure out yet why making them optional's not working, I know that our current `shouldComponentUpdate` in the `PlotlyEditor` would have to be adjusted to make this work. But even with an adjustment there, something's not quite right and my app is stuck in an infinite loop instead of rendering. Browser just crashes.

How do you find the `README` for the lib?

bpostlethwaite:
which README lib?

bpostlethwaite:
also can you file a bug about the revision numbers?

README.md
```diff
@@ -5,60 +5,60 @@
...
+     props:
+     * data, layout, revision: from application state
+     * onUpdate: the handlePlotUpdate callback created above
+   * Editor component: is imported from `react-plotly.js-editor`, it requires
+     these props:
+     * dataSources, dataSourceOptions, graphDiv, revision: from application
+       state
+     * onUpdate: the handleEditorUpdate callback above
+     * plotly: the plotly.js library
+     * locale: more on this later
```
bpostlethwaite: locals isn't required right? Optional?

VeraZab:
@bpostlethwaite merging this one, alright?

bpostlethwaite:
:dancer:

Issue #157, opened 2017-12-01
CSS makes some big assumptions on absolute positioning. 


bpostlethwaite:
First attempt to throw the simplest Plotly.js + Editor into a create-react-app app.

![image](https://user-images.githubusercontent.com/1176674/33505696-6c3613b6-d6ba-11e7-8693-0a933897e2a4.png)

The Panels have no set width and are positioned `100px` from the right... so if you have a margin...

bpostlethwaite:
I believe this one is addressed now right @VeraZab ?

bpostlethwaite:
Sounds like you have made some updates too @aulneau. What if we edit one of our examples such that the editor appears _within_ a more complicated nested structure? Right now all of our sample apps allow it to take over the screen.

aulneau:
@bpostlethwaite Yeah that sounds like a good idea! I'll whip some examples up. Are there any use cases you know of where it would exist in a nested layout, I was thinking a modal could make sense.

bpostlethwaite:
Sure yes that is a good idea. Another common case will be the canonical bootstrap dashboard layout. nav header and left nav side-panel that toggles inner content views one of which will be the editor and plot.

aulneau:
Cool okay, perfect. I was thinking it might make sense to allow people to change the order of `sidebar` and `panel`. In situations where people might have the editor on the right, we could allow for the sidebar to be flush against the right of the container. But that might be a sep issue.

bpostlethwaite:
yeh could be a sep issue. Another related but separate issue is that I would like a way to programmatically change which panel is open. Specifically some way of putting the bit of state that controls the panel into the Application state rather than the editor state. But yeh I'll make an issue for that specifically.

Issue #138, opened 2017-11-29
Fancy scroll bar


bpostlethwaite:
![image](https://user-images.githubusercontent.com/1176674/33377845-e45c195c-d4e0-11e7-9918-eed1130a4112.png)

PR #190, opened 2017-12-27
Design updates

### LINKED ISSUE #171: required import of react-select.css isn't great
Perhaps we can incorporate the css or host the css in our scss build system? It might be a good idea to incorporate any file copies into the css build step.

aulneau:
Okay so this addresses a number of things:

Updated styling for the sidebar & panel (includes scrolling when needed for both sidebar and panel):
sidebar

I've changed the panel and sidebar from using floats and absolute positioning to a flex based layout. I have also made a lot of updates to make sure things are spaced evenly and consistently (eg: items in the panel, .fold + .fold, etc).

I've added `babel-plugin-module-resolver` so that we can import things as `import {bem} from 'lib'` vs `import {bem} from '../../../lib'`. If we don't want to do that, just lmk and I can remove it!

I refactored the Fold component a bit to be a bit more 'componentized'.

I added a `Panel.Header` sub-component to the `Panel` component. I have noticed a pattern of having actions in the head of the panel, and so I think we should use a component for that header that takes an action, and in the future could take a title/label and potentially additional actions (eg: 'collapse all' in a hidden menu or something)

I've added `mdi-react` which allows us to import icons: `import CloseIcon from 'mdi-react/CloseIcon` in place of a font-icon. I prefer this method of icons as they are svgs that get embedded directly in the component. You can see them in use [here](https://github.com/plotly/react-plotly.js-editor/compare/master...design-updates#diff-59516a8f55e012e8d54b19b9d109e105).

I've also gone through and started improving the scss around, converting hex's to variables and trying to replace anything we use more than once with variables (borders, text-shadows, etc).

This fixes #171 #138 #157


bpostlethwaite:
We use eslint and prettier src code formatting and we have tests in CI that check for this. There is bound to be prettier hooks for Webstorm - I use the autoformat on save feature. Could you get those CI tests passing then ping me again, thanks!

src/components/containers/Panel.js
```diff
@@ -1,21 +1,45 @@
...
+    if (visible) {
+      return <div className={bem('panel')}>{children}</div>;
     }
-    return <div />;
+    return null;
   }
 }
 
+Panel.Header = PanelHeader;
+
```
VeraZab: why not just export PanelHeader itself? we don't use this convention anywhere else.
some files have multiple exports, we export each individually, and import with `{}`

src/components/containers/__tests__/Fold-test.js
```diff
@@ -36,7 +36,7 @@ describe('<Fold>', () => {
         </TraceFold>
       </TestEditor>
     )
-      .find('.fold__delete')
+      .find('.js-fold__delete')
       .simulate('click');
```
VeraZab: nice thanks!

src/styles/components/widgets/_multi-format-text-editor.scss
```diff
@@ -113,7 +185,7 @@ $multi-format-editor-background: #fafbfd;
   box-shadow: none;
 
   // add our styles
-  border-top: 1px solid $color-border;
+  //border-top: 1px solid $color-border;
   padding: 6px;
```
VeraZab: delete?

src/styles/components/widgets/_multi-format-text-editor.scss
```diff
@@ -212,12 +284,20 @@ $multi-format-editor-background: #fafbfd;
...
+  border: $border;
+  padding: $half-spacing-unit;
   position: absolute;
-  width: 235px;
+  //width: 268px;
+  //height: 115px;
+  //left: auto !important;
+  //top: 195px !important;
+  //transform: translateX(-1px);
   z-index: 1;
```
VeraZab: delete?

utils/combined-translation-keys.txt
```diff
@@ -158,5 +158,5 @@ q1:
 q3:                                                                                                         // plotly.js: traces/box/calc.js:131
 source:                                                                                                     // plotly.js: traces/sankey/plot.js:140
 target:                                                                                                     // plotly.js: traces/sankey/plot.js:141
-trace                                                                                                       // plotly.js: plots/plots.js:439
+trace                                                                                                       // plotly.js: plots/plots.js:442
 upper fence:                                                                                                // plotly.js: traces/box/calc.js:135
```
VeraZab: oh right, the css changes change the line references of the translation strings, great, thanks for updating!

package.json
```diff
@@ -20,10 +20,11 @@
     "make:dist": "mkdirp dist && browserify src/PlotlyEditor.js -o ./dist/PlotlyEditor.js -t [ babelify --presets [ es2015 react ] ] -t browserify-global-shim --standalone createPlotlyComponent && uglifyjs ./dist/PlotlyEditor.js --compress --mangle --output ./dist/PlotlyEditor.min.js --source-map filename=dist/PlotlyEditor.min.js.map && make:dist:css",
     "make:dist:css": "mkdirp dist && node-sass --output-style compressed src/styles/main.scss > dist/react-plotly.js-editor.css",
     "prepublishOnly": "npm run make:lib && npm run make:combined-translation-keys",
-    "lint": "prettier  --write \"src/**/*.js\"",
-    "test:lint": "eslint src && echo -e '\\033[0;32m'PASS'\\033[0m'",
+    "prepublishOnly:scss": "npm run make:lib:css && npm run make:combined-translation-keys",
+    "lint": "prettier --write \"src/**/*.js\"",
```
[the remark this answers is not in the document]
VeraZab: Got it, translation keys line references change as the css gets updated and js refactored, thanks!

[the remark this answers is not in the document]
aulneau: Oh great -- I just added this after talking with @bpostlethwaite, didn't realize watched worked :D thanks! will remove

src/components/containers/Panel.js
aulneau: I did it mostly for backwards compatibility and it's a convention I recently started following for creating compound components. For example, I can go into the Batch app and just use `<Panel.Header>...</Panel.Header>` without having to import an additional thing. I didn't want to change the convention of panel from `import Panel from...` to `import {Panel, PanelHeader} from...` as that would break every instance of `import Panel`. We could do `import Panel, {PanelHeader} from...`.

VeraZab: > We could do import Panel, {PanelHeader} from....
Yeah, it's the one we use throughout our code I think

src/components/containers/Fold.js
```diff
@@ -1,6 +1,8 @@
 import PropTypes from 'prop-types';
 import React, {Component} from 'react';
 import classnames from 'classnames';
+import CloseIcon from 'mdi-react/CloseIcon';
+import ChevronDownIcon from 'mdi-react/ChevronDownIcon';
```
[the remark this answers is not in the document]
bpostlethwaite: @VeraZab you don't think that it looks a lot better process than having to go into a 3rd party webapp, select an icon, download it, run some hashing on the font, add the font to the editor, reference it with a class and then link to it in the code?

src/styles/components/widgets/_multi-format-text-editor.scss
bpostlethwaite: @aulneau yep in general we delete rather than comment-out. The code will be in git history anyway. Otherwise they accumulate over time and it's hard to know why it was commented out.

Issue #174, opened 2017-12-19
Editor initializes trace as "line" though chart type selector says "scatter"


bpostlethwaite:
Ok well it's a "scatter+line" but we should probably treat a "scatter" as a scatter to avoid confusion.

![image](https://user-images.githubusercontent.com/1176674/34161093-72923334-e49d-11e7-8db3-7f783649843f.png)

nicolaskruchten:
I had wondered about this in the past :)

PR #191, opened 2017-12-28
Trace name fix

VeraZab:
fixes: #174

### DIFF
--- a/src/lib/customTraceType.js
+++ b/src/lib/customTraceType.js
@@ -1,4 +1,4 @@
-export function plotlyToCustom(trace) {
+export function plotlyTraceToCustomTrace(trace) {
   if (
     trace.type === 'scatter' &&
     ['tozeroy', 'tozerox', 'tonexty', 'tonextx', 'toself', 'tonext'].includes(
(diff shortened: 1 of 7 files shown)


VeraZab:
alright @bpostlethwaite change of plan :) will make small prs that address the official release milestone as priority. 

So here, just a few explanations, when sending plotly.js a 'scatter' type without a mode, plotly.js will automatically assume it's a marker+lines mode, that's why in the trace recognition function I added 'marker+lines' to the 'lines' trace type

Yup, let me know if for any other clarifications, thanks!

src/lib/customTraceType.js
```diff
@@ -0,0 +1,32 @@
+export function plotlyToCustom(trace) {
```
bpostlethwaite: `plotlyToCustomTrace` and `customTraceToPlotly`? Not a big deal but it isn't apparent from the function name what it is actually doing.

bpostlethwaite: (could be custom axes or custom sandwiches :) )

bpostlethwaite: or even more explicit `plotlyTraceToCustomTrace` and `customTraceToPlotlyTrace` ? This ought to appeal to someone who likes typing `MultiTextEditor` instead of `TextEditor` ;)

VeraZab: haha, alright!

VeraZab:
💃 ?

bpostlethwaite:
:dancer:

Issue #184, opened 2017-12-26
Consistent way to override CSS and styling


bpostlethwaite:
We have style/variables/main.scss which exposes variables users can override to configure the style of the editor. However these styles are consistently applied within the editor. They are not really consistent themselves. 

We need to push more variables into "themable" component scss and expose a more consistent style "API"

I'll have to discuss what is actually required for theming but I think we can go pretty far until I find that info out.

aulneau:
I'm thinking about this... some other components / ui libraries that I've used in the past allow for you to import something like a `ThemeProvider` and then you can pass an obj of style overrides to the component. I think that _might_ work well, I'll have to think about it some more. I also need to refactor a lot of the scss to allow for things to be consistent and abstract enough for everything to work out. Will continue to update this thread as I progress.

aulneau:
So there are a couple options I am thinking about.

If the editor will be _mostly_ used in a sandboxed chromium environment, then I think css variables would be a great way to allow for theming.

For example:

```
:root {
--brand-color: #bababa;
--button-primary-fill: #another
}
```

and then the user could override the styles by including a new definition after they have included the components styles

```
import react-plotly.js-editor/lib/react-plotly.js-editor.css
import styles/plotly.js-editor-custom-theme.css
```

and the `plotly.js-editor-custom-theme.css` could contain one or many new variable definitions that would be applied throughout the application.

The only issue with using css custom props is that they are _not_ super supported:

![screen shot 2017-12-28 at 12 07 20 pm](https://user-images.githubusercontent.com/11803153/34419405-541935d2-ebc9-11e7-9a7e-e4b8bc28057d.png)

Another option would be to use something like [this](https://github.com/javivelasco/react-css-themr), which could also be used if we ever move to [styled-components](https://github.com/styled-components/styled-components)

bpostlethwaite:
The editor will be mostly used in Plotly's Chart Studio. Next will be random installs and apps that depend on the open source lib. But working in a chrome embedded environment is of course a hard requirement.

bpostlethwaite:
The only concerning item in that support document is IE 11 which we support in the Chart Studio. If we can work around that by having the defaults === Chart Studio defaults then that shouldn't be an issue?

aulneau:
[...]

Issue #183, opened 2017-12-26
CSS - allow multiple widgets in widget field


bpostlethwaite:
Adding this commit cfd7acb broke they styling of Multiple Values
![image](https://user-images.githubusercontent.com/1176674/34362085-6e5dca3e-ea3e-11e7-8c36-5d0d43bf6fc6.png)

bpostlethwaite:
Needed for an application that adds multiple items (looks like this: 
![image](https://user-images.githubusercontent.com/1176674/34362110-9801c534-ea3e-11e7-807e-54fdf9875ee3.png) )

aulneau:
@bpostlethwaite Do you think it makes sense to add that remove functionality you have for the group by fields inside of the editor? I like this display and interactivity of adding multiple fields, it reminds me of adding stops on the uber app.

Do we also want 'Multiple Values' to the select grouping columns section?

bpostlethwaite:
Change it up as you want. We just need to be able to add and remove fields.

I prefer individual rows of groups rather than a single dropdown that can contain multiple values. IMO we want to look fairly standard when built into a native desktop app and not look so web 3.0 as to stick out. But it can and should look great.

aulneau:
fixed via https://github.com/plotly/react-plotly.js-editor/pull/195

Issue #76, opened 2017-11-15
zero state handling


bpostlethwaite:
If there are no traces with data the panels should show something useful or friendly. See WS2 handling.

aulneau:
fixed via https://github.com/plotly/react-plotly.js-editor/pull/195

PR #195, opened 2017-12-31
Styles updates, Theming and other misc updates

aulneau:
Okay, I've gone through just about every style in this thing and updated / refactored them to be cleaner and use variables whenever possible.

I've also added a new step to our css, I've added postcss to process our generated css and do a few different things. It seemed like we weren't autoprefixing our css at all, so I've added `autoprefixer`. I've also opted to use `cssnano` with postcss to minify our css in that process, so we are no longer using the `node-sass compress` functionality. I've added a postcss plugin to combine duplicate selectors into one.

For IE11, I've added a couple postcss plugins to convert all of our custom css-props to their true values, so that if people need to support IE11 (eg: chart studio), you just include the ie11 stylesheet instead. This step can also be integrated into the users app, so that if they wanted to theme the editor first, and then run the postcss plugin themselves, they could. I've also added a plugin to remove `:root{}` from the stylesheet if you are generating styles for ie11. 

I've also added an empty state for if there is no trace data, the message could/should be modified:

screen shot 2017-12-31 at 4 35 42 pm

I have added a delete component within the field component:

screen shot 2017-12-31 at 4 36 27 pm

I've also added a Button component.

I still need to go through and update some icons to what we have in `plotly-icons`, but I wanted to get this PR open :)

Regarding theming: I've opted to use scss maps and then parse through them to generate our scss variables.

The various scss maps:

screen shot 2018-01-02 at 10 53 06 am

are mapped through with scss functions:

screen shot 2018-01-02 at 10 53 18 am

which then are outputted as css variables:

screen shot 2018-01-02 at 10 52 00 am

which can be easily overridden by the end user in their app:

screen shot 2018-01-02 at 10 54 29 am

which allows for a themable app:

screen shot 2018-01-02 at 10 55 15 am

resolves: #184 #183 #76


bpostlethwaite:
Awesome work! I haven't looked in depth at this yet and maybe @VeraZab can do a review of the CSS? 
As part of this PR could you add language to the README that tells people how they would customize this library? That and the build process are what I will focus on (the style and build API for developers). 

Also AFAIK you are not restricted to NODE_ENV. You could, for example, use another env variable called PLOTLYJS_EDITOR_CSS_TARGET and then set that in the build command before running the javascript that does the building. 

NODE_ENV is usually used to distinguish between 'dev' and 'production' builds.

bpostlethwaite:
Also this PR seems to address more than one existing issue. I see "empty state for not traces". Could you update your description to include "resolves #x #y and #z" type declarations? Github will auto-close the issues and it will give me more context. Thanks!

package.json
```diff
@@ -79,14 +79,20 @@
...
+    "browserify-global-shim": "^1.0.3",
+    "cssnano": "^3.10.0",
     "draft-js": "^0.10.4",
     "draft-js-export-html": "github:plotly/draft-js-export-html",
     "draft-js-import-html": "^1.2.1",
     "draft-js-utils": "^1.2.0",
+    "es6-shim": "^0.35.3",
     "fast-isnumeric": "^1.1.1",
+    "mdi-react": "^2.1.19",
     "plotly-icons": "^1.0.1",
```
VeraZab: @aulneau you're going to add the additional mdi-react icons we need to plotly-icons?

src/components/containers/TraceAccordion.js
```diff
@@ -27,9 +29,13 @@ export default class TraceAccordion extends Component {
...
 
     const addButton = canAdd && (
-      <button className="panel__add-button" onClick={this.addTrace}>
-        + Trace
-      </button>
+      <Button
+        className="js-add-trace-button"
+        variant="primary"
+        onClick={this.addTrace}
+        icon={<PlusIcon />}
```
VeraZab: or PlusIcon could be the default in the Button component?

VeraZab: or maybe not, if we're including modals to the editor (which I'm not sure we will be), maybe we want more flexibility on that button

src/components/sidebar/SidebarGroup.js
```diff
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React, {Component} from 'react';
-import {bem} from '../../lib';
-
+import {bem} from 'lib';
+import ChevronRightIcon from 'mdi-react/ChevronRightIcon';
 import SidebarItem from './SidebarItem';
```
VeraZab: it's AngleDownIcon in plotly-icons

src/styles/_helpers.scss
```diff
@@ -1,3 +1,33 @@
...
+.\+hover-grey:hover {
+  color: $color-active-text;
+}
+
+@function spacing($name, $true-val:false) {
+  @if $true-val == true {
+    @return map-get($spacings,$name);
+  } @else {
+    @return var(--spacing-#{$name});
+  }
```
VeraZab: is this a sass thing: `var(--spacing-#{$name})` how does it work?

src/styles/_mixins.scss
```diff
@@ -34,7 +34,7 @@
   }
 }
 
-@mixin scrollbar($width: 5px, $trackbg: $color-page-background, $thumb-color: $color-accent) {
+@mixin scrollbar($width: 5px, $trackbg: var(--scrollbar-track-background), $thumb-color: var(--scrollbar-thumb-color)) {
   -webkit-overflow-scrolling: touch;
```
VeraZab: what's the difference here between `$color-accent` and `var(--scrollbar-thumb-color)`

aulneau: So a lot of these variables are kind of passed around so that you can theme the application, the reason this is set this way is so that the end user can change the accent color, and the scollbar default thumb color would be the accent color, but if they wanted a different color for the scrollbar, they would change the value of `--scrollbar-thumb-color`.

src/styles/_helpers.scss
aulneau: whoops, these are no longer used, I'll remove them. 

I've added a lot of scss maps and parsers, to out put css custom properties (native css variables) from our scss.

src/components/containers/TraceAccordion.js
aulneau: Maybe... I think it might be better to leave no default icon, because there are buttons that don't have icons.

src/components/sidebar/SidebarGroup.js
aulneau: Cool, I'll make sure to update all of the icons throughout with our `plotly-icons` and add any I use throughout :)

aulneau:
Yeah I know :/ I felt so bad making like 6 more scripts in there... There has to be a way to abstract and make it more efficient.

src/components/containers/TraceAccordion.js
```diff
@@ -40,8 +46,23 @@ export default class TraceAccordion extends Component {
...
 
+    const emptyState = data.length &&
+      !data[0].x &&
+      !canAdd && (
+        <PanelEmpty
+          heading="There aren't any traces."
+          message={
+            <p>
+              Looks like there aren't any traces defined yet. Go to the 'Create'
+              tab to define some traces.
```
bpostlethwaite: this needs to be wrapped in i18n `_()`. Ping me or @VeraZab with any questions

src/components/containers/TraceAccordion.js
```diff
@@ -27,9 +29,13 @@ export default class TraceAccordion extends Component {
...
     const addButton = canAdd && (
-      <button className="panel__add-button" onClick={this.addTrace}>
-        + Trace
-      </button>
+      <Button
+        className="js-add-trace-button"
+        variant="primary"
+        onClick={this.addTrace}
+        icon={<PlusIcon />}
+        label="Trace"
```
bpostlethwaite: this needs to be wrapped in i18n

src/components/containers/AnnotationAccordion.js
```diff
@@ -30,14 +31,18 @@ export default class AnnotationAccordion extends Component {
...
     const addButton = canAdd && (
-      <button className="panel__add-button" onClick={this.addAnnotation}>
-        + Annotation
-      </button>
+      <Button
+        variant="primary"
+        className="js-add-annotation-button"
+        onClick={this.addAnnotation}
+        icon={<PlusIcon />}
+        label="Annotation"
```
bpostlethwaite: i18n

src/components/Button.js
```diff
@@ -0,0 +1,44 @@
+import PropTypes from 'prop-types';
```
bpostlethwaite: Button should go in `components/widgets`?

src/styles/components/_button.scss
```diff
@@ -0,0 +1,73 @@
+// button default styles
```
bpostlethwaite: this should be within a folder in components/? Seems like a widget to me.

aulneau:
Okay @bpostlethwaite and @VeraZab, I've updated the readme, added a theming document, and I believe I've fixed everything you two have requested. If you can review my latest changes, I'd say this is good to go! :) (barring any new changes requested)

VeraZab:
yey 💃

package.json
```diff
@@ -13,27 +13,29 @@
...
     "make:translation-keys": "babel-node utils/findTranslationKeys.js",
-    "make:lib":
-      "mkdirp lib && babel src --out-dir lib --ignore=__tests__/* --source-maps && npm run make:lib:css",
-    "make:lib:css":
-      "mkdirp lib && node-sass src/styles/main.scss > lib/react-plotly.js-editor.css",
-    "make:dist":
-      "mkdirp dist && browserify src/PlotlyEditor.js -o ./dist/PlotlyEditor.js -t [ babelify --presets [ es2015 react ] ] -t browserify-global-shim --standalone createPlotlyComponent && uglifyjs ./dist/PlotlyEditor.js --compress --mangle --output ./dist/PlotlyEditor.min.js --source-map filename=dist/PlotlyEditor.min.js.map && make:dist:css",
-    "make:dist:css":
-      "mkdirp dist && node-sass --output-style compressed src/styles/main.scss > dist/react-plotly.js-editor.css",
+    "make:lib": "mkdirp lib && babel src --out-dir lib --ignore=__tests__/* --source-maps && npm run make:lib:css",
```
bpostlethwaite: @VeraZab I think you and @aulneau have squabbling .json formatting. I don't think we should use prettier on anything outside `/src` or at least not on anything other than *.js files.

bpostlethwaite:
thanks for the heroic level of effort here! Can't wait to try out some new themes! 

:dancer:

VeraZab:
new release after this merge! : )

package.json
aulneau: Yeah, sometimes my muscle memory causes me to autoformat anything after I hit save. But it's strange, my current package.json is formatted correctly:

screen shot 2018-01-03 at 12 03 03 pm

aulneau:
@VeraZab okay I fixed the `prepublishOnly` task. should be good to go! (also I don't have permission to merge this PR)

Issue #70, opened 2017-11-15
Info message for when trace opacity is not supported


bpostlethwaite:
Area charts for example do not support trace level opacity. See workspace2 info message.

VeraZab:
I don't get it, isn't it even better that this styling option doesn't even show up in the editor, since that chart type doesn't have it?
![screen shot 2018-01-02 at 9 37 09 pm](https://user-images.githubusercontent.com/4257572/34507557-3afb5540-f005-11e7-8c7e-e4b2a862a257.png)

bpostlethwaite:
We have it in Chart Studio and that is our specification. Someone at some point put some thought into it and decided it was a good idea. I don't know who it was so I can't argue the point but that is why I am trying to replicate what we already have and not generate regressions.

bpostlethwaite:
That doesn't mean _all_ the ideas were good ideas but you should probably ask CP and AJ or Chels. Someone probably dealt with a confused customer at one point.

alexcjohnson:
I suspect the message is there because it's often not obvious *why* the opacity option has disappeared - see the logic involved in https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/clean_data.js - it can even be the properties of a *different* trace that invalidate opacity for this trace.

alexcjohnson:
FWIW, scatter is currently the only trace type that has a `module.cleanData` method allowing traces to affect each other's attributes. I'll make a mental note that if we ever need to make another `cleanData` it may warrant another explanatory note to users!

PR #198, opened 2018-01-03
Add info message for area trace opacity

VeraZab:
closes: #70 
screen shot 2018-01-03 at 12 21 11 pm

### DIFF
--- a/src/components/containers/Section.js
+++ b/src/components/containers/Section.js
@@ -2,10 +2,13 @@ import Info from '../fields/Info';
 import MenuPanel from './MenuPanel';
 import React, {Component, cloneElement} from 'react';
 import PropTypes from 'prop-types';
-import unpackPlotProps from 'lib/unpackPlotProps';
-import {containerConnectedContextTypes} from 'lib/connectToContainer';
+import {
+  containerConnectedContextTypes,
+  localize,
+  unpackPlotProps,
+} from '../../lib';
 
-export default class Section extends Component {
+class Section extends Component {
   constructor(props, context) {
     super(props, context);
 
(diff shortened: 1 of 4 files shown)


src/components/containers/Section.js
```diff
@@ -21,12 +24,23 @@ export default class Section extends Component {
...
 
+    /*
+     * show opacity disabled message for scatter traces that have a fill 'tonexty' || 'tonextx'
+     * based on:
+     * https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/clean_data.js#L13
+     */
+    const cannotSetTraceOpacity =
+      fullContainer.type === 'scatter' &&
+      (fullContainer.fill !== 'none' || fullContainer.fill !== 'toself');
+
```
VeraZab: @alexcjohnson is `cannotSetTraceOpacity` correct?

I feel like since there's an accordion menu for each trace in the editor, the rest of this logic doesn't apply: 
https://github.com/plotly/plotly.js/blob/master/src/traces/scatter/clean_data.js#L24

alexcjohnson: if we're looking at the trace from `fullData`, we should just be able to use the results of `scatter/clean_data`, ie `fullContainer.type === 'scatter' && fullContainer.opacity === undefined`, right? Because the multi-trace aspect of it is really important for alleviating confusion and you're right that you don't want to try and reconstruct that here.

VeraZab: thanks! will do that

VeraZab:
@bpostlethwaite @alexcjohnson can I merge?

alexcjohnson:
> can I merge?

want to rebase while I look at it again?

PR #203, opened 2018-01-04
Account for long text

aulneau:
This PR is to account for longer text in most areas. The only spot that I think could cause issue is on the sidebar. It's pretty fixed, and until we have actual examples of length of words, I am not sure what we should do about that. We could allow for larger sidebar based off of a language. 

This pr will prevent text from overflowing in the title area of folds, adding an '...' to the end of the string when it starts to overflow:

screen shot 2018-01-04 at 2 12 18 pm

This also passes className's down to the fold if you apply them on the component.


VeraZab:
or text-font could get smaller? for languages who's menus translate to really long words?

VeraZab:
but 💃  for this pr

aulneau:
@VeraZab yeah ideally it'd be great to not adjust type size, as we're already pretty small as it stands. Most areas (fold titles, buttons, section headers, field labels) all work with long text as I tested it (they will just wrap and look ok). The sidebar is fixed at a small width, though, so wrapping won't really be possible (esp with 1 word long items).

alexcjohnson:
> so wrapping won't really be possible (esp with 1 word long items).

can we hyphenate aggressively? you're right that sidebar titles really do NOT want to be truncated...

Issue #153, opened 2017-11-30
Collapse all


nicolaskruchten:
Copy feature from Workspace

VeraZab:
ok, starting this one

VeraZab:
Hmm, yeah, for TraceAccordion and Annotations accordion, Folds are not children..
unless there's something I'm not seeing..

bpostlethwaite:
Hmmm what if Accordions are HOCs around Panels? That would change the DefaultEditor and other apps that use the Panels and Accordions but it isn't too late at this point. Actually that wouldn't work for cases like: 
```
<Panel>
   <AnnotationAccordion />
   <Fold />
</Panel>
```
where we wouldn't want `<AnnotationAccordion>` to be a panel itself.

Okay what if the Panel checks for `Accordion` types and then digs into the children of those as well? `Accordion` types could pass props from `Panel` to `Fold` . There is some degree of coupling here though. 

@aulneau you have any thoughts on this?

aulneau:
hmm I wonder if `context` could help here? If panel could contain a collapse function or something in its `context` and then in the fold component we could look for that context, rather than having to traverse / deep look into the children.

bpostlethwaite:
The only issue here is that the Panel needs to know if there are Fold children and how many. If there are none it won't display the Collapse Icon. Ditto for 1. Only when there are 2 or more Folds will the Panel show the collapse button. 2 way information in a 1 way data flow problem... never an elegant solution. 

I guess the least bad is just counting children and because we have Accordions the children of child Accordions...

bpostlethwaite:
Though we could still pass the `toggleFold` function through context. that way there wouldn't be Accordions handing down props... not sure.

bpostlethwaite:
If we are going to root up to 2 levels of children and do something special with Accordions we already have hierarchical constraints 
```
Panel
    Fold
```
or
```
Panel 
    Accordion
        Fold
```
So also coupling together with props doesn't seem so bad. If the components are rendered inside other things we can't count the children and then doesn't matter that the chain of props is also broken.

PR #204, opened 2018-01-05
Make Accordions collapsible

VeraZab:
closes: #153 
 
ex


VeraZab:
This pr makes panels smart, they detect whether or not folds are present, and make a collapse all button appear if more than one visible fold is present.

This assumes that if we're going to use folds within a panel, a foldIndex prop will have to be present on all our folds, so that each fold can look up its state from the array of fold states that is passed to it via context from the Panel component. Passing fold states from context makes it work for both panels that have folds as children directly and panels that have accordions that render folds.

@nicolaskruchten or @alexcjohnson or @bpostlethwaite would you review please

src/components/containers/Fold.js
```diff
@@ -4,21 +4,44 @@ import classnames from 'classnames';
 import {CloseIcon, AngleDownIcon} from 'plotly-icons';
 
 export default class Fold extends Component {
-  constructor() {
-    super();
-    this.state = {folded: false};
+  constructor(props, context) {
+    super(props, context);
+    const {individualFoldStates} = context;
+    if (!individualFoldStates) {
```
nicolaskruchten: Could we just make `individualFoldStates` required and stop supporting both modes? Is there a use-case for a Fold that's not in a Panel like this?

src/components/containers/Panel.js
```diff
@@ -1,51 +1,139 @@
...
+    };
+  }
 
+  toggleFolds() {
+    const {individualFoldStates} = this.state;
+    const hasOpen =
+      individualFoldStates.length > 0 &&
+      individualFoldStates.some(s => s !== true);
+    this.setState({
+      individualFoldStates: individualFoldStates.map(() => {
```
nicolaskruchten: This could just be `.map( () => hasOpen )` right?

src/components/containers/PanelHeader.js
```diff
@@ -0,0 +1,79 @@
...
+      <div className="panel__header">
+        {children && children.length ? (
+          <div className="panel__header__content">{children}</div>
+        ) : null}
+        <div className="panel__header__actions__container">
+          {allowCollapse ? (
+            <div className="panel__header__collapse" onClick={toggleFolds}>
+              {hasOpen ? (
+                <span>
+                  <ResizeDownIcon />
```
nicolaskruchten: The cursor on these collapse/expand buttons is not currently a pointer cursor, but it should be.

src/components/containers/AnnotationAccordion.js
```diff
@@ -25,51 +16,43 @@ class AnnotationAccordion extends Component {
...
 };
 
 AnnotationAccordion.propTypes = {
-  localize: PropTypes.func,
   children: PropTypes.node,
   canAdd: PropTypes.bool,
 };
 
-export default localize(AnnotationAccordion);
+export default AnnotationAccordion;
```
nicolaskruchten: Why are we removing the localization here?

src/components/containers/Panel.js
```diff
@@ -1,51 +1,139 @@
...
+        if (hasOpen) {
+          return true;
+        }
+        return false;
+      }),
+    });
+  }
+
+  toggleFold(index) {
+    const copy = [...this.state.individualFoldStates];
```
nicolaskruchten: this would be a good candidate for https://github.com/kolodny/immutability-helper

src/components/containers/AnnotationAccordion.js
VeraZab: because we no longer have any strings to localize in this component :)

src/components/containers/Fold.js
VeraZab: There's not a use case yet, just thought it might have been useful to not have this coupled.

src/components/containers/PanelHeader.js
VeraZab: oh yeah, thanks!

src/components/containers/Fold.js
nicolaskruchten: In that case my (weak) preference would be to couple them strongly by making the props and contexts required, so as to avoid people occasionally forgetting to add them everywhere etc. YAGNI :)

VeraZab: Yeah, I think you're right, I'll do that

src/components/containers/Panel.js
VeraZab: I'll add this one, I think it'll be helpful, thanks!

nicolaskruchten: Maybe would be worth doing that in a single pass where we refactor other places where we have used various spread/copy tricks, rather than adding this dependency just for this case. I just wanted to advocate for this library because I've found it so, so useful in many little places :)

nicolaskruchten: Especially useful with Redux, as Ben used it in the batch repo

VeraZab: hmm, well, for now, I'm just going to add it in this one place, or at least, this pr is only going to change it in this one place :) I'll open an issue for changing it throughout the editor.

Yeah, I've thought I've seen Ben use something to help with this, but I didn't find anything in the repo, it was in the batch one! thanks!

nicolaskruchten: OK, sounds good!

src/components/containers/Panel.js
```diff
@@ -1,51 +1,141 @@
...
+        return false;
+      }),
+    });
+  }
+
+  toggleFold(index) {
+    this.setState(
+      update(this.state, {
+        individualFoldStates: {
+          [index]: {$set: !this.state.individualFoldStates[index]},
```
nicolaskruchten: Here I would use `$toggle` instead of `$set` like this: `update(this.state, {individualFoldStates: {$toggle: [index]}})`

VeraZab: oh, thanks :)

nicolaskruchten: I know the docs say that `$toggle` works on objects but this works, I just tried it

VeraZab:
yeah, now, I need to adjust tests to have that individualFoldStates array in context...

VeraZab:
yeah..I'm having a bit of difficulties on those Panel tests.
I'd like to merge this in as is, and add tests later.

@nicolaskruchten would you sign off on this one, I know that it's not ideal without tests, but I'd like for us to have all of our base features in first.

nicolaskruchten:
retroactive 💃

PR #205, opened 2018-01-05
Add plot type icons to trace fold.

aulneau:
This PR adds icons for plot types in the trace fold.

plot-type-icons

This also abstracts out the widths of the editor, sidebar and panel a bit better to allow for customizing via theming / overrides.


src/lib/connectTraceToPlot.js
```diff
@@ -42,6 +46,7 @@ export default function connectTraceToPlot(WrappedComponent) {
         container: trace,
         fullContainer: fullTrace,
       };
+      this.type = plotlyTraceToCustomTrace(trace);
```
aulneau: @bpostlethwaite // @VeraZab let me know if you think there is a better place for this to live

aulneau:
^ I believe the tests are failing because I am using icons that are not currently in the `plotly-icons` npm package.

src/components/containers/Fold.js
```diff
@@ -50,10 +58,32 @@ export default class Fold extends Component {
...
+        case 'line':
+          return <PlotLineIcon />;
+        case 'bar':
+          return <PlotBarIcon />;
+        case 'pie':
+          return <PlotPieIcon />;
+        case 'area':
+          return <PlotAreaIcon />;
+        default:
+          return <PlotScatterIcon />;
```
VeraZab: can we make the default icon something else but scatter?
there's maybe 20-ish trace types in plotly.js, that makes for a long switch statement, could we write this in a more dynamic way..

src/styles/components/fields/_field.scss
```diff
@@ -70,7 +70,7 @@
       opacity: 1;
       svg {
         path {
-          fill: var(--color-sienna);
+          //fill: var(--color-sienna);
         }
```
VeraZab: erase it instead of commenting out?

src/lib/connectTraceToPlot.js
VeraZab: I think its fine here

VeraZab:
ah i see, you were npm linked and that's why it was working for you ..
yeah, I think once we merge and release all the new plot-icons, these errors should go away

src/styles/components/fields/_field.scss
aulneau: oh whoops! missed this -- was just testing when I wasn't sure where that red was coming from. Will fix.

src/components/containers/Fold.js
aulneau: I know.. 

I wonder if something like this is possible -

```
const iconName = `Plot${c(type)}Icon`;
return React.createElement(iconName, null, null);
```

c(string) capitalizes the first letter of the type.

bpostlethwaite: It's pretty easy to get all the Plotly traces loaded in a particular instance. See https://github.com/plotly/react-plotly.js-editor/blob/master/src/components/fields/TraceSelector.js#L15 for an example.

Or from a console with a Plotly on window do something like 
```
var schema = Plotly.PlotSchema.get()
console.log(Object.keys(schema.traces))
```

bpostlethwaite: (for example https://plot.ly/create)

aulneau: yeah I am thinking about implementing something like:

```
import Icons, {otherIcons} from 'plotly-icons'

...

const renderPlotIcon = (type) => {

    const iconName = `Plot${c(type)}Icon`;

    return Icons[iconName] || defaultIcon;

}

```

Thoughts?

VeraZab: I think its a good idea to have a map of icon names that link to the components.
And I think that map should be a constant in the react-plotly.js-editor

That map could be generated through Plotly.PlotSchema.get(), like Ben pointed out, and then we could dynamically check if we have a an icon for that plotly.js plot type:
require('plotly-icons')['PlotScatterIcon'] // we get the icon
require('plotly-icons')['PlotScatterIco'] // undefined

If we don't have an icon, we'd use a default one.

bpostlethwaite: What would the map do that 
```
const iconName = `Plot${c(type)}Icon`;
```
isn't?

bpostlethwaite: If the plot-icons have a strict format for using the trace names and we have the trace names in editor then 
```
 const iconName = `Plot${c(type)}Icon`;

 return Icons[iconName] || defaultIcon;
```
would work right?
Then if plotly.js adds a new trace type (happens every other month) we add the icon in plot-icons and that's it (aside from bumping the plot-icon version). No need to modify a map in the editor.

VeraZab: I'm talking about how we'd generate `Icons`, I'm saying it should be a variable in the editor instead of an export in plotly-icons.
by map I mean that `Icons = {'PlotBarIcon': <PlotBarIcon/>}`
and that it can be dynamic, like you said, but a constant built out in the editor.

bpostlethwaite: I don't totally follow. Can you post a pseudo-code snippet that imports the icons from plotly-icons and constructs the `Icon` object?

VeraZab:
Alright, now Icons are loaded dynamically, as for me I think this pr is good to go!

VeraZab:
@bpostlethwaite what do you think of this one now?
although, now that I think of it more, that map of plot icons should probably be computed once, in PlotlyEditor, and then passed down as context

src/lib/index.js
```diff
@@ -25,14 +26,40 @@ function getDisplayName(WrappedComponent) {
...
+}
+
 const TOO_LIGHT_FACTOR = 0.8;
 export function tooLight(color) {
   const hslColor = tinyColor(color).toHsl();
   return hslColor.l > TOO_LIGHT_FACTOR;
 }
 
+function getPlotIcons(plotly) {
+  const icons = Object.keys(plotly.PlotSchema.get().traces).reduce(
```
bpostlethwaite: we already pass the plotSchema down in the context so no need to call `.get()` here. See the editor.

re: your comment about computing this once... that seems reasonable.

bpostlethwaite: wait a sec though

bpostlethwaite: Do we need to iterate over all the traces? Presumably when we need to show an icon we already have the trace type.

bpostlethwaite: I would rather not store an object if we don't need it.

src/components/containers/Fold.js
```diff
@@ -79,7 +83,9 @@ Fold.propTypes = {
   children: PropTypes.node,
   hideHeader: PropTypes.bool,
   name: PropTypes.string,
+  type: PropTypes.string,
   className: PropTypes.string,
+  plotIcons: PropTypes.object,
```
bpostlethwaite: the Fold should be given an icon rather than a `type` and a big map of icons.

src/lib/connectTraceToPlot.js
```diff
@@ -71,7 +79,14 @@ export default function connectTraceToPlot(WrappedComponent) {
     }
 
     render() {
-      return <WrappedComponent name={this.name} {...this.props} />;
+      return (
+        <WrappedComponent
+          name={this.name}
+          type={this.type}
+          plotIcons={this.plotIcons}
```
bpostlethwaite: What if the wrapped component isn't a `<Fold>` ?

package.json
```diff
@@ -75,6 +76,7 @@
     "gl": "^4.0.4",
     "glob": "^7.1.2",
     "jest": "^21.2.1",
+    "jest-cli": "^22.0.4",
```
bpostlethwaite: why is this needed?

src/lib/connectTraceToPlot.js
bpostlethwaite: That could be ok so long as we have a consistent API for accepting icons in our components. Then any `icon` aware component can be swapped in here are it'll work. Do we have any other components that take `icon` as props? If so lets get the naming standardized.

bpostlethwaite:
```
➜  src git:(TemporaryFixForAxesTypeClearing) ✗ ack PropTypes | grep icon
components/containers/Panel.js:66:  icon: PropTypes.node,
components/containers/MenuPanel.js:58:  iconClass: PropTypes.string,
components/widgets/Button.js:41:  icon: PropTypes.any,
components/widgets/RadioBlocks.js:72:      icon: PropTypes.string,
```

bpostlethwaite: MenuPanel takes an `iconClass`. Seems like it needs to be updated to our new pattern.

package.json
aulneau: I was getting test errors saying it was not available, I added that and it worked -- but I can remove it and see if my tests fail.

src/lib/index.js
```diff
@@ -36,25 +36,11 @@ export function tooLight(color) {
...
-
-  // We have to add some of the editor specific trace types as they're not
-  // in plotly.PlotSchema.get()
-  icons.line = PlotlyIcons.PlotLineIcon;
-  return icons;
+function renderTraceIcon(trace) {
+  const componentName = `Plot${capitalize(trace)}Icon`;
+  return PlotlyIcons[componentName]
+    ? PlotlyIcons[componentName]
+    : PlotlyIcons.PlotLineIcon;
```
aulneau: okay @bpostlethwaite @VeraZab how does this look?

src/lib/connectTraceToPlot.js
aulneau: it looks like `MenuPanel` isn't using iconClass at all, we can probs just remove it.

package.json
VeraZab: I can't run npm run test:js without it

src/lib/index.js
VeraZab: but aren't there some folds that are rendered and don't belong to a trace? so then the icon should just be null?

VeraZab: but yeah, otherwise, looks good! that's the idea!

aulneau: yeah so this function should only be called on the traceFold area (or anywhere we want to generate a specific trace icon), so we want to render a trace icon but if the naming is wrong or the icon doesn't exist, we will render the default trace icon (which we can change if `PlotLineIcon` isn't what we want as a default. The fold takes a `icon` prop and if the icon is passed it will render, otherwise it wont.

package.json
bpostlethwaite: How have we been running tests for the past few months? It sure has been working for me. Did it get removed perhaps?

VeraZab: well... i was cheating, installing it locally on my branch everytime, and this time I just got tired of doing it..

bpostlethwaite: I guess it's because I always run tests like `../node_modules/jest/bin/jest` instead of using `npm run test:js` :P  ? Well our npm scripts ought to work so :+1:

bpostlethwaite: Though how does it work in Circle CI? Puzzling.

aulneau:
So this is super close -- but my tests are failing and I don't know why :c 

screen shot 2018-01-08 at 11 50 21 am

VeraZab:
ok, I'll take a look

src/components/containers/MenuPanel.js
```diff
@@ -32,7 +38,7 @@ export default class MenuPanel extends Component {
 
   render() {
     const {show, ownline, label, children} = this.props;
-    const isOpen = show || this.state.isOpen;
+    const {isOpen = show} = this.state;
```
bpostlethwaite: @VeraZab @aulneau the logic here has changed. Before we would take `show` first and then `isOpen` after. Now `state.isOpen` takes precedence. 

What that means is that the `show` props is effectively useless because `this.state.isOpen` is always defined.

This is the source of the test failure

aulneau: Whoops you're totally right! that's my bad, I changed that line. Thanks @bpostlethwaite

bpostlethwaite: The previous logic `show || this.state.isOpen;` is accurate. If `show` is `false` then defer to the component state. If show is `true` then the component will always be open.

aulneau:
okay @bpostlethwaite and @VeraZab this should be good to go.

I've tacked on a few other changes within this PR:

mentions styles
editor with abstraction (of variable values)
Icons for trace types (+ icons as elements throughout)

bpostlethwaite:
@aulneau should `mentions` be in the Editor? It isn't an editor component. Just a one off usage.

src/components/containers/Fold.js
```diff
@@ -49,15 +49,14 @@ export default class Fold extends Component {
...
+
+    const icon = Icon && <Icon className="fold__top__icon" />;
 
     const foldHeader = !hideHeader && (
       <div className={headerClass} onClick={this.toggleFold}>
         <div className="fold__top__arrow-title">
           {arrowIcon}
-          {plotTypeIcon}
+          {icon}
           <div className="fold__top__title">{name}</div>
```
VeraZab: you know, I've found that when we don't explicitly return a null, in the event that `Icon && <Icon className="fold__top__icon" />;` is undefined, the `{icon}` in foldHeader will be rendered by react as a `0`.

At least that's what happens in our accordion menus now, since we're not returning null for these: https://github.com/plotly/react-plotly.js-editor/blob/master/src/components/containers/TraceAccordion.js#L69

They turn into a `zero` when we delete all traces. 

so could you have `icon ? icon : null` or some other variant up there, just to make sure.

for TraceAccordion, I fixed these in my collapse pr.

src/components/widgets/RadioBlocks.js
```diff
@@ -39,12 +39,8 @@ class RadioBlocks extends Component {
...
         onClick={() => this.handleChange(value)}
       >
-        {icon ? (
-          <span className="radio-block__icon">
-            <i className={icon} />
-          </span>
-        ) : null}
-        {label ? <span>{label}</span> : null}
+        {Icon && <Icon className="radio-block__icon" />}
+        {label && <span>{label}</span>}
```
VeraZab: yup, here too, can we explicitly render null

src/components/containers/Fold.js
```diff
@@ -84,7 +84,7 @@ Fold.propTypes = {
   name: PropTypes.string,
   type: PropTypes.string,
   className: PropTypes.string,
-  icon: PropTypes.element,
+  icon: PropTypes.any,
 };
```
VeraZab: why are we changing it back to any?
I would have put icon: PropTypes.component

aulneau: Well we're technically passing a function as an icon for the trace folds.

screen shot 2018-01-08 at 1 01 41 pm

I haven't used `PropTypes.component` before, would that account for the function, too?

src/components/containers/Fold.js
aulneau: I'll go through and remove those types of conditionals and replace with an explicit `null`

src/components/containers/Fold.js
bpostlethwaite: If the function is a class constructor for a react component then `Proptypes.node` should work. (I think that is what @VeraZab is referring to there is no `.component` AFAIK.)

aulneau: Yeah I tried that -- still get `Failed prop type: Invalid prop `icon` supplied to `Fold`, expected a ReactNode.`

How about `icon: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),`?

bpostlethwaite: man, ok whatever works

aulneau:
okay @VeraZab I believe I updated all of the things you requested :)

VeraZab:
thank you, great 🎉

Issue #211, opened 2018-01-09
make empty state for each individual trace


VeraZab:
like on plot.ly/create
![screen shot 2018-01-09 at 8 58 25 am](https://user-images.githubusercontent.com/4257572/34724293-4b03b2d2-f51b-11e7-9ecc-547e270318dc.png)

PR #214, opened 2018-01-09
Add trace empty state

VeraZab:
closes: #211

screen shot 2018-01-09 at 4 29 54 pm

### DIFF
--- a/src/components/containers/Fold.js
+++ b/src/components/containers/Fold.js
@@ -2,8 +2,9 @@ import PropTypes from 'prop-types';
 import React, {Component} from 'react';
 import classnames from 'classnames';
 import {CloseIcon, AngleDownIcon} from 'plotly-icons';
+import {localize} from 'lib';
 
-export default class Fold extends Component {
+class Fold extends Component {
   render() {
     const {deleteContainer, individualFoldStates, toggleFold} = this.context;
     const {
@@ -14,6 +15,8 @@ export default class Fold extends Component {
       foldIndex,
       canDelete,
       icon: Icon,
+      isEmpty,
+      localize: _,
     } = this.props;
 
     const folded = individualFoldStates[foldIndex];
@@ -63,9 +66,20 @@ export default class Fold extends Component {
       </div>
     );
 
-    const foldContent = !folded ? (
-      <div className={contentClass}>{children}</div>
-    ) : null;
+    let foldContent = null;
+    if (!folded && !isEmpty) {
+      foldContent = <div className={contentClass}>{children}</div>;
+    }
+
+    if (!folded && isEmpty) {
+      foldContent = (
+        <div className={contentClass}>
+          {icon}
+          <div>{_('This trace does not yet have any data.')}</div>
+          <div>{_('Return to the Graph > Create menu above to add data.')}</div>
+        </div>
+      );
+    }
 
     const classes = className ? ' ' + className : '';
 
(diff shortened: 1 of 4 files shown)


VeraZab:
@aulneau would you review this one please

src/components/containers/Fold.js
```diff
@@ -63,9 +66,26 @@ export default class Fold extends Component {
...
+            <div className="fold__content__empty__item">{icon}</div>
+            <div className="fold__content__empty__item">
+              {_('This trace does not yet have any data.')}
+            </div>
+            <div className="fold__content__empty__item">
+              {_('Return to the Graph > Create menu above to add data.')}
+            </div>
+          </div>
+        </div>
+      );
```
aulneau: Hmm, but what if a fold needs an empty state, but an icon is not passed?

```
 <div className="fold__content__empty__item">{icon ? icon : defaultIcon}</div>
```

I would maybe abstract this empty fold item out as something like 

```
<EmptyFoldContent
    show={!folded && isEmpty}
    icon={icon}
    message={['This trace does not yet have any data.', ''Return to the Graph > Create menu above to add data.']} />
```
What do you think @VeraZab ?

aulneau: Yeah maybe these messages: 'This trace does not yet have any data.' should not live in the `Fold` component, but be in the instance of the fold that needs the message?

VeraZab: yeah you're right, I'll make a separate more generic component and I'll put a default icon

VeraZab:
what it looks like
screen shot 2018-01-09 at 8 35 33 pm

I think this covers your comments, would you take another look @aulneau

src/components/containers/FoldEmpty.js
```diff
@@ -0,0 +1,30 @@
...
+      <div className="fold__content__empty">
+        <div className="fold__content__empty__icon">
+          {Icon ? <Icon /> : null}
+        </div>
+        <div className="fold__content__empty__message__primary">
+          {messagePrimary ? messagePrimary : null}
+        </div>
+        <div className="fold__content__empty__message__secondary">
+          {messageSecondary ? messageSecondary : null}
+        </div>
```
aulneau: I tend to prefer to have the conditional on the `<div...` itself, because often the div's will have padding on them or some other layout modifier, and we'd want to remove that if the content is gone, too

src/components/containers/Fold.js
```diff
@@ -63,9 +66,22 @@ export default class Fold extends Component {
...
+      foldContent = <div className={contentClass}>{children}</div>;
+    }
+
+    if (!folded && isEmpty) {
+      foldContent = (
+        <div className={contentClass}>
+          <FoldEmpty
+            icon={Icon}
+            messagePrimary={isEmpty.messagePrimary}
+            messageSecondary={isEmpty.messageSecondary}
```
aulneau: My only concern here is having two props for messages (What if we need three or more?), versus passing a message prop that could be either a single node or an array of strings/components, eg:

pseudo code --

```
const checkType = (message) => {
    if (typeof message === 'array') {
        message.map((messageItem, i) => <div key={i} className="message">{message}</div>)
    } else {
        return (<div className="message">{message}</div>);
    }
}

message={[_('localized string 1), _('localized string 2)]} || message={_('localized single string)} || message={<SingleComponent />}
```

src/styles/components/containers/_fold.scss
```diff
@@ -113,6 +113,48 @@
...
+      }
+
+      &__message__secondary {
+        font-family: var(--font-family-body);
+        font-size: var(--font-size-small);
+        text-align: center;
+        font-size: 13px;
+        color: var(--color-text-base);
+      }
+    }
```
aulneau: I would probably have`&__message__title` class and then `&__message__body` (for nth body items)

src/components/containers/Fold.js
VeraZab: yeah, we could envision this as we extend the functionality, for now I think it'll work ok

aulneau:
But otherwise this looks good!

PR #216, opened 2018-01-10
Adjust read me

VeraZab:


examples/simple/package.json
```diff
@@ -7,7 +7,6 @@
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
     "react-plotly.js": "^1.0.4",
-    "react-plotly.js-editor": "0.2.0-alpha.1",
```
nicolaskruchten: why does the `simple` app not get its own editor but the others do?

VeraZab: oh, I think this must have happened when npm link was ran.

I added an `npm setup` script to the editor that does the npm linking of the editor to the simple app, 

so that when someone clones the editor and just wants to see how it works or get up and running developing, they can just clone, npm setup, npm start, like Ben did in his batch repo

it's when the setup script runs that it must have been removed, I'll add it back

nicolaskruchten: OK. We could also have the setup script not do the `npm link` dance, though, right?

VeraZab:
empty panel look after the changes
screen shot 2018-01-10 at 10 57 12 am

examples/simple/package.json
VeraZab: yeah, i could remove that

package.json
```diff
@@ -20,7 +20,7 @@
     "make:lib:js": "mkdirp lib && babel src --out-dir lib --ignore=__tests__/* --source-maps",
     "make:lib:css": "mkdirp lib && babel-node scripts/styles.js && SASS_ENV=ie babel-node scripts/styles.js && babel-node scripts/postcss.js && SASS_ENV=ie babel-node scripts/postcss.js",
     "prepublishOnly": "npm run make:lib",
-    "setup": "npm install && npm link && rm -rf node_modules/react && rm -rf node_modules/react-dom && cd examples/simple && npm install && npm link react-plotly.js-editor",
+    "setup": "npm install && cd examples/simple && npm install && cd ../redux && npm install && cd ../async-data && npm install",
```
nicolaskruchten: Ah OK if we're doing things this way, then we could consider skipping `setup` and adding these as a post-install hook

VeraZab: hmm, you know I was just adding this in so that people could have an easy command to run to get an example running and play around with the code a little. 

So thinking back of the original motivations, I think I'll remove the npm installs to the redux and async-data examples, and keep that npm link to the simple repo.

Issue #217, opened 2018-01-10
Axes title input does not reset when adding initial input


bpostlethwaite:
![image](https://user-images.githubusercontent.com/1176674/34789992-47e7f8bc-f60e-11e7-9114-cb36937b0a88.png)

It should have placeholder functionality

PR #222, opened 2018-01-10
Add placeholder value to titles

VeraZab:
closes: #217 
ex

### DIFF
--- a/src/components/widgets/text_editors/MultiFormatTextEditor.js
+++ b/src/components/widgets/text_editors/MultiFormatTextEditor.js
@@ -209,7 +209,14 @@ class MultiFormatTextEditor extends Component {
       return null;
     }
 
-    const {onChange, placeholder, value, localize: _} = this.props;
+    const {onChange, localize: _} = this.props;
+    let {placeholder, value} = this.props;
+
+    if (value.startsWith('Click to enter') && value.endsWith('title')) {
+      placeholder = `${value.replace('Click to enter', 'Enter')}`;
+      value = '';
+    }
+
     const {currentTab} = this.state;
 
     const richTextClassNames = classnames(
--- a/src/components/widgets/text_editors/MultiFormatTextEditor.js
+++ b/src/components/widgets/text_editors/MultiFormatTextEditor.js
@@ -209,13 +209,7 @@ class MultiFormatTextEditor extends Component {
       return null;
     }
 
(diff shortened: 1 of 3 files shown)


src/components/widgets/text_editors/MultiFormatTextEditor.js
```diff
@@ -209,7 +209,14 @@ class MultiFormatTextEditor extends Component {
       return null;
     }
 
-    const {onChange, placeholder, value, localize: _} = this.props;
+    const {onChange, localize: _} = this.props;
+    let {placeholder, value} = this.props;
+
+    if (value.startsWith('Click to enter') && value.endsWith('title')) {
```
VeraZab: or @alexcjohnson would you know of a better way for me to get that default string when title's not set?

VeraZab:
ready for review @alexcjohnson @bpostlethwaite

src/components/widgets/text_editors/MultiFormatTextEditor.js
alexcjohnson: Mmm, that way isn't going to work with i18n anyway. Is it possible to compare the value from `fullLayout` with the value from `layout`? If they're different, I think that's a robust signal that you have the placeholder. And if you explicitly typed in the placeholder, then it's not really a placeholder anymore :)

alexcjohnson: If we're not comfortable using the plotly.js placeholder verbatim, I think we'd have to explicitly make and localize a new placeholder for each attribute that uses this. Doesn't seem worth it to me, frankly; I'd either use the one from plotly.js or use a standard placeholder for every text entry box.

VeraZab: yeah, we could use those plotly.js strings verbatim, it's easier for localization too, like you mention, so I'll put those strings as placeholders then, not values

VeraZab:
ex

container is giving me that attribute's immediate container, so this seems to work ok to me
@alexcjohnson ?

VeraZab:
oups, I forgot to also check against attr
so `attr === 'title' && !container.title`, seems to be what I need to check against

alexcjohnson:
Great! I'm happy with this, thanks for iterating through it with me. 💃

VeraZab:
thanks!

PR #225, opened 2018-01-11
make and write an arrows translation we can use for testing

alexcjohnson:
This is the "arrows" transform Alteryx uses to test translations. It appears to be based on estimates that German (and Dutch, apparently) tend to be about 35% longer when translated from English, and they do it by padding with unicode arrows, to both test unicode support and to make it obvious if any strings have not been wrapped appropriately.

Dunno if this PR really belongs here or in batch - anyway the code can probably be ported over directly, along with find/combine keys to include those in batch. @bpostlethwaite @VeraZab thoughts?


VeraZab:
I think it could live in this repo, like translation-key scripts, and its a nice tool for ui testing @aulneau what do you say?

aulneau:
oh cool, I think this would be totally useful! I'm not sure about where it should live either, but I think it would be fine for it to live here, too.

alexcjohnson:
Chatting with @bpostlethwaite I think we want this here AND in batch, to support various workflows. So I'll merge this and then port the framework over to batch too. @VeraZab look OK now? `prettier` is happy finally.

VeraZab:
looks great 💃

Issue #73, opened 2017-11-15
Documentation


bpostlethwaite:
Investigate options including https://github.com/documentationjs/documentation

PR #230, opened 2018-01-12
Documentation

nicolaskruchten:
First pass at #73 within a README rather than with something like Styleguidist, which can come later if needed.


README.md
```diff
@@ -71,6 +89,106 @@ To get around this, you can create an
...
+* `<ColorPicker />`: renders as a popup color-picker, useful for CSS color hex value strings
+* `<Flaglist />`: renders as a list of checkboxes, useful for `+`-joined flag lists like `data[].mode`
+* `<MultiFormatTextEditor />`: renders as a WYSIWYG editor, useful for text like `layout.title`
+* `<Info />`: renders as text, useful for displaying help text (unbindable)
+
+<p align="center">
+  <img src="examples/custom/components.png" alt="Components" width="432" height="692" border="1">
+</p>
+
+### Connector functions
```
bpostlethwaite: I'm tempted to move `Connector functions` to the bottom of this section. Connector functions are only needed when building your own containers which can be considered "advanced". 
Or perhaps we have `standard usage` and `advanced usage` sections. 
Just thinking out loud.

README.md
```diff
@@ -71,6 +89,106 @@ To get around this, you can create an
...
+cd examples/simple # or any other application directory
+npm install
+rm -rf node_modules/react node_modules/react-dom node_modules/react-plotly.js-edit
+npm link react-plotly.js-edit
+npm start # keep this running
+```
+
+## Built-in Components
+
+This module provides a number of nestable _containers_ which are intended to contain _fields_ that render _widgets_ that have been connected to individual values in the figure via _connector functions_. Containers can also be connected to certain parts of the figure tree such that their child fields map to the appropriate leaf values. A field must have a connected container as an ancestor in order to be bound to the figure. The `<PlotlyEditor />` and connector functions use the React [`context` API]() to push configuration information to child components.
```
bpostlethwaite: This is a pretty heavy paragraph. I see you discuss bound and unbound components below so maybe we should keep this here but perhaps we can try and draw a line between "typical" usage of the library vs "advanced" (rolling your own components). 

We don't want users to freak out about HOCs and context when all they are looking for is a Field that will let them change the colour of the plot.

Though I suppose basic usage is just the `DefaultEditor`. Perhaps this is fine.

examples/README.md
```diff
@@ -0,0 +1,6 @@
+# `react-plotly.js-editor` examples
+
+* [Simple `react-plotly.js-editor` example](simple): `DefaultEditor`, synchronous data, top-level component state management
+* [Async-data `react-plotly.js-editor` example](async-data): `DefaultEditor`, asynchronous data, top-level component state management
```
bpostlethwaite: fwiw now that we don't need an evented system I think we should change this example to use `fetch` and `promises` instead of events. That will be useful to a wider percentage of users. Not necessary in this PR.

examples/custom/src/registerServiceWorker.js
```diff
@@ -0,0 +1,108 @@
+// In production, we register a service worker to serve assets from local cache.
```
bpostlethwaite: you can :zap: this file and all other service worker files from our examples (and of course the import of these files). They're only useful in production environments.

README.md
```diff
@@ -18,43 +20,59 @@ npm start
...
-       state
-     * onUpdate: callback that updates state new plotRevision number
-     * plotly: the plotly.js library
-     * locale: if using the default locale 'en', it is not necessary to pass in
-       this prop, more on locales later
-
-## Styling
+## Overview
+
+This module's entry point is a React component called `<PlotlyEditor />` which connects to a [Plotly.js](https://plot.ly/javascript/)-powered `<Plot />` component care of [`react-plotly.js`](https://github.com/plotly/react-plotly.js). A Plotly.js plot is defined by a JSON-serializable object called a _figure_. `<PlotlyEditor />` accepts as children React components whose descendents are input elements wrapped via `connectToContainer()` calls so as to bind them to the `<Plot />`'s figure's values. If no children are passed to the `<PlotlyEditor />`, the `<DefaultEditor />` is used. This module also exposes the [building block components](#Built-in-Components) that comprise the `<DefaultEditor />` so that developers can create their own customized editors.
```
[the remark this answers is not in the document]
jackparmer: Let's put a "Screenshots & GIFs" section of what this thing is before the "To consider" section. Could be just 1 screenshot or GIF. https://github.com/apache/incubator-superset does a good job of this.

bpostlethwaite:
fyi we try and rebase away commits like "tweak" and "wip". Either squash them together or squash them into previous commits if the changes are just tweaks on earlier commits.

nicolaskruchten:
Thanks for the feedback! Re the tweak/wip stuff, I almost always "squash and merge" so that none of these commits land on master, no worries :)

README.md
```diff
@@ -71,6 +89,106 @@ To get around this, you can create an
...
+npm link       # registers react-plotly.js-editor for subsequent npm link calls
+npm run watch  # keep this running
+```
+
+In another terminal:
+
+```
+cd examples/simple # or any other application directory
+npm install
+rm -rf node_modules/react node_modules/react-dom node_modules/react-plotly.js-edit
```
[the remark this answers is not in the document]
nicolaskruchten: I think this is OK for now.

README.md
nicolaskruchten: Yes, it is pretty heavy but I've found it helpful to have such a thing in the docs somewhere: a straightforward statement of the architecture that people can refer to and say "yep, that about covers it". It's not for everyone but some people appreciate it. 

I do think that basic usage is the `DefaultEditor` and that anything beyond that will need to get to this level of detail. Certainly I was not able to build a simple editor without it. If you try to build an editor with just a Field to change a colour, you will need to have *at least* a layout-connected container, which you must (currently) manually construct out of `connectLayoutToPlot()` and `Panel` (because we don't export e.g. `LayoutPanel`). We can probably make that easier by exporting more stuff out of the box which might lighten up the docs, but I figured I'd start by documenting what we had and then iterating to simplify :)

README.md
nicolaskruchten: See above: until we export some of the pre-connected components like `LayoutPanel` and `AxesFold` I think we are stuck with documenting these somewhat like this, otherwise people can't build all that much.

examples/README.md
nicolaskruchten: Agreed.

examples/custom/src/registerServiceWorker.js
nicolaskruchten: Agreed.

README.md
nicolaskruchten: (we can discuss in #237)

nicolaskruchten:
Addressed all feedback, ready for rereview :)

bpostlethwaite:
Nice work! :dancer: 

For the record @nicolaskruchten and I discussed exposing preconnected containers like `LayoutPanel` to make some of the extra complexity unnecessary. But we can do that in a later PR.

PR #240, opened 2018-01-15
correct iteration over children

nicolaskruchten:
Fix for #228 ... uses `React.Children.map` to iterate over children without assuming it's an array.

/cc @VeraZab for future reference https://mxstbr.blog/2017/02/react-children-deepdive/


VeraZab:
thanks @nicolaskruchten !

PR #246, opened 2018-01-17
Share i18n

alexcjohnson:
- Generalizes the translation scripts @VeraZab made so we can reuse them in batch (didn't work for `findTranslationKeys` due to the different babel versions... but the other two are shared)
- Localizes a few more things that were missed earlier

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,9 @@
   },
   "scripts": {
     "lint": "prettier --write \"src/**/*.js\"",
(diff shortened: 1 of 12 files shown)


alexcjohnson:
@VeraZab can you give this a quick review, then help me publish it so I can use it in https://github.com/plotly/plotly.js-batch-reporter/pull/24 ?

VeraZab:
yeah, was looking at it, well, looks good, so you'd like a new patch release?

alexcjohnson:
> so you'd like a new patch release?
yes please 🙏

PR #248, opened 2018-01-17
add afterTracesUpdate hook

bpostlethwaite:
@nicolaskruchten turns out the app I am working on needs a gd.data update event _after_ the update. That way I can collect all the next sources rather than figure out which of the previous sources will be updated and generating a new list (this is exactly what the editor does in its update routine).

Having a _before_ hook is useful. Do you think we should we add `after` hooks for every update method or introduce these as necessary?


bpostlethwaite:
Okay ready for review @nicolaskruchten !

Issue #215, opened 2018-01-10
Empty state for annotations when no traces present


nicolaskruchten:
In the existing workspace this is what appears when no traces exist:

![image](https://user-images.githubusercontent.com/203523/34751342-04d1d1ae-f57a-11e7-9914-c3ebb8d1f2d2.png)

PR #250, opened 2018-01-17
Language style fixes, empty state for style panels

### LINKED ISSUE #241: Legend Panel Zero state
When there are no traces it makes little sense to allow legend configuration.

aulneau:
This commit makes some adjustments to the layout for longer words in different languages. I made the confirmation overlay scrollable, reduced type size in field labels and in radio blocks, and made an adjustment to the dropdown placeholder ('...' was being added because it already had '...' on it). I also modified the modal box to take up slightly more space.

screen shot 2018-01-17 at 1 57 28 pm

screen shot 2018-01-17 at 1 57 39 pm

I also have implemented a check for if a panel requires traces, and if there are no axis set with data, it will render an empty state.

I did notice however that if you set an axis, and then delete the whole trace (hitting the X on the trace fold) the layout data is not purged, and so the application still thinks that there is at least one trace. cc @bpostlethwaite 

resolves #241 #215


src/components/containers/Panel.js
```diff
@@ -103,7 +105,18 @@ export default class Panel extends Component {
...
+              <PanelEmpty
+                heading={_("Looks like there aren't any traces defined yet.")}
+                message={<p>{_("Go to the 'Create' tab to define traces.")}</p>}
+              />
+            ) : (
+              newChildren
+            )
+          ) : (
+            newChildren
+          )}
```
alexcjohnson: This can be simplified - `requireTraces && !hasTraces ? <PanelEmpty /> : newChildren`

But also, looks like `<PanelEmpty>` is only called here and in `<TraceAccordion>`, with the same translated strings both times - can we just do those once inside `<PanelEmpty>`?

src/components/widgets/Dropdown.js
```diff
@@ -64,7 +64,7 @@ class Dropdown extends Component {
       <div className={dropdownContainerClass} style={dropdownStyle}>
         <Select
           backspaceToRemoveMessage={backspaceToRemoveMessage}
-          placeholder={placeholder || _('select an option...')}
+          placeholder={placeholder || _('Select an Option')}
```
alexcjohnson: I don't see these in the key files - can you collect them again?
FWIW it's best to do `npm run make:combined-translation-keys && npm run make:arrows` - the former is included in `make:lib` but not the latter, we should probably add it. We don't actually use `xx.js` from this repo right now, we only use the one created in batch, but eventually when we're using this editor in other contexts we probably will use the one here.

nicolaskruchten:
Are there any `LayoutPanel`s without `requireTraces`? Should we just bake that into `LayoutPanels` ?

aulneau:
@nicolaskruchten that's what I was thinking, but I was not quite sure the best way to bake it into `LayoutPanel`!

nicolaskruchten:
@aulneau maybe have the current `LayoutPanel` get renamed to `LayoutConnectedPanel` and not exported and then export a simple `LayoutPanel` component that renders as a `LayoutConnectedPanel` with the `PanelEmpty` logic there rather than in `Panel` ? Would be nice to stop putting so much specificity in `Panel` if possible... We already inspect the children to see if they're `TraceAccordions` or whatnot ;)

alexcjohnson:
> Are there any `LayoutPanel`s without `requireTraces`? Should we just bake that into `LayoutPanels` ?

Wait, is that what we want? I'd have thought you could still monkey with the size, axes, and annotations even with no traces... but you need traces to see the legend.

nicolaskruchten:
This is basically just cloning the behaviour of the workspace: you can't monkey with sizes/axes without a trace there at the moment.

nicolaskruchten:
Sorry, hit wrong button!

alexcjohnson:
> you can't monkey with sizes/axes without a trace there at the moment.

I see. OK, well whether or not we want to do it that way in the `<DefaultEditor>` (and I kind of think we don't but that's a discussion to have), lets not bake it directly into `LayoutPanel`, because it might not be what everyone building an editor wants.

nicolaskruchten:
OK. Can we make a `TraceRequiredPanel` then, so that this stuff doesn't end up in `Panel`? It seems like `Panel` should be as generic as possible and specialized elsewhere.

aulneau:
Sure thing, I'm headed out for a bit but I can make that when I am back tonight.

aulneau:
Okay! this should be good to go again :)

aulneau:
Actually @nicolaskruchten, when I merged in the latest updates from `master`, my local app is now breaking -- the data seems to be not be saving anymore in the batch app. I've been looking through as I see there was a bump to the `react-plotly.js` you did in the example apps. I bumped the versions in the batch app on a local branch, but it's still not working. Everything had been working before updating to the latest from master.

src/components/containers/TraceRequiredPanel.js
```diff
@@ -0,0 +1,76 @@
...
+    const {hasTraces} = this.state;
+
+    if (visible) {
+      /**
+       * Check if there is any trace data
+       */
+      if (
+        Object.keys(layout.xaxis).length === 0 ||
+        Object.keys(layout.yaxis).length === 0
+      ) {
```
VeraZab: hmm, I think it's better to look if there's traces by going into _fullData.
there's a key in each trace, called visible, so I think something like this should be pretty accurate:
fullData.filter(trace => trace.visible).length > 0

aulneau:
Okay! this is ready for (hopefully) a final review :) cc @bpostlethwaite @VeraZab

src/components/containers/TraceRequiredPanel.js
```diff
@@ -12,17 +15,14 @@ class TraceRequiredPanel extends Component {
...
     if (visible) {
       /**
        * Check if there is any trace data
        */
-      if (
-        Object.keys(layout.xaxis).length === 0 ||
-        Object.keys(layout.yaxis).length === 0
-      ) {
+      if (fullData.filter(trace => trace.visible).length === 0) {
         if (hasTraces) {
```
VeraZab: I would have put `fullData.filter(trace => trace.visible).length === 0 && hasTraces`, but it's ok :)

Issue #255, opened 2018-01-18
some way to add buttons to Editor Sidebar


bpostlethwaite:
We need this for the workspace integration and I need it for the batch app.

![image](https://user-images.githubusercontent.com/1176674/35114063-8136baf6-fc51-11e7-831d-8a247eeb6a23.png)

Should we make it easy to build a custom Sidebar or should the editor have an API for this or what?

PR #256, opened 2018-01-18
Button in menu

VeraZab:
to close: #255 
@bpostlethwaite what would you think of something like this?
@aulneau if that principle's ok, would you help me with the styling of this...

screen shot 2018-01-18 at 3 13 45 pm


bpostlethwaite:
nice! yeh this declarative API seems to be consistent with how we do everything else. Can you add a button to the Redux example that fires some redux action?

src/components/PanelMenuWrapper.js
```diff
@@ -53,16 +56,28 @@ class PanelsWithSidebar extends Component {
       group = child.props.group;
       name = child.props.name;
 
-      if (groupLookup.hasOwnProperty(group)) {
-        groupIndex = groupLookup[group];
-        obj = groups[groupIndex];
-      } else {
+      if (child.type.name === 'Button') {
```
bpostlethwaite: Are we expecting this always to be our `Button` type? If so we can compare classes instead of child.type.name. From Section.js

```
if (child.type === Info) {
...
}
```

Do we need to restrict users to our `Button`? Is there a way to allow users to pass any arbitrary component to be added to the Sidebar.

What if we had a `SidebarItem` component that wraps whatever you want to include in the sidebar?

```jsx
<PanelMenuWrapper >
   <SidebarItem name="save">
       <a onClick={saveMyApp} />
    </SidebarItem>
</PanelMenuWrapper>
```
This way users can put any thing in a `SidebarItem` container and we just render it.

src/components/PanelMenuWrapper.js
```diff
@@ -53,16 +56,28 @@ class PanelsWithSidebar extends Component {
...
+        if (groupLookup.hasOwnProperty(group)) {
+          throw new Error('cannot have 2 menu buttons with same name');
```
bpostlethwaite: Is there a solid reason why we can't have 2 buttons with the same name?

src/components/widgets/Button.js
```diff
@@ -18,7 +26,13 @@ class Button extends Component {
       classes += ` button--default`;
     }
 
-    classes += ` ${className}`;
+    if (group) {
+      classes += ' button--menu';
```
bpostlethwaite: In the name of keeping our widgets dumb perhaps instead of adding logic that knows about groups in a the button widget maybe we should just pass in a className. So th parent would supply `button--menu` and the child would simply render whatever class it is passed.

src/components/PanelMenuWrapper.js
```diff
@@ -53,16 +56,28 @@ class PanelsWithSidebar extends Component {
...
+          component: child,
+        };
         groups.push(obj);
+      } else {
+        if (groupLookup.hasOwnProperty(group)) {
+          groupIndex = groupLookup[group];
+          obj = groups[groupIndex];
+        } else {
+          groupLookup[group] = groups.length;
+          obj = {name: group, panels: []};
```
bpostlethwaite: Could `obj` be `{name: group}` with no panels? Then later on we would handle the case where `panels` is undefined?

not a huge deal but having to create empty arrays seems to suggest we generalize the framework just a bit.

src/styles/variables/_theme.scss
```diff
@@ -88,6 +88,7 @@ $fold: (
 );
 // Buttons
 $buttons: (
+  menu:(padding-top: 10px),
```
aulneau: hmm I would perhaps not do this, and instead use one of the spacing variables we have eg, `padding-top: var(--spacing-quarter-unit);` on the `.button--menu` class, so that we can keep all spacing units consistent

src/styles/components/widgets/_button.scss
```diff
@@ -51,7 +51,9 @@ button {
       padding-left: 0;
     }
   }
-
+  &--isMenu {
+    @include button(menu);
```
aulneau: This looks like it might not actually be doing anything. I don't see any edits to the mixin for this, `@mixin button{ ... }` found in the `_mixins.scss` partial.

VeraZab:
ok, so looks like this now:
screen shot 2018-01-19 at 12 36 45 am

VeraZab:
alright, @bpostlethwaite @aulneau ready for another look!

src/components/widgets/index.js
```diff
@@ -0,0 +1,2 @@
+import Button from './Button';
```
bpostlethwaite: I hadn't added a widget index since widgets are meant to be internal. That said I am using some in the batch app so apparently they are not that internal. I suppose we may as expose more widgets!

bpostlethwaite: we could put Dropdown in here eventually as well (and other useful components folks may wish to use in their apps)

examples/custom/src/CustomEditor.js
```diff
@@ -69,6 +71,20 @@ export default class CustomEditor extends Component {
             </Section>
           </Fold>
         </LayoutPanel>
+        <SingleSidebarItem>
+          <Button
```
bpostlethwaite: Yeh I like this API!

bpostlethwaite:
thanks for putting this in looks great :dancer: !

Issue #258, opened 2018-01-19
More robust interaction with children


nicolaskruchten:
In a few places in the code we have some components that behave differently based on the nature of their children, which makes sense but is a bit brittle at the moment, because HOCs work by composition, not inheritance. For example we can't reliably just check the `type` of the child or do an equality-check on its `displayName` because that will generally fail when the child is wrapped. 

The safest thing I've found so far is to do an `indexOf` check on the `displayName` after statically declaring it so that minification won't mangle it. Clearly this is a bit of a hack :)

I wonder if something a bit more robust/explicit/duck-typing-friendly isn't in order, something like an `editorTraits` object that components would statically declare and that `connect*` method would respect so that a parent could check if a child had a specific trait instead of checking some approximation of its identity. 

I'll do some research on how others have tackled this problem with HOCs and report back, but now is a good time to try to improve our system, before we start piling on more special cases :)

nicolaskruchten:
Another way to do this would be just with abitrarily-named static properties, and to follow the 'official' HOC recommendation to always pass statics along: https://reactjs.org/docs/higher-order-components.html#static-methods-must-be-copied-over

This would be helpful: https://github.com/mridgway/hoist-non-react-statics

PR #262, opened 2018-01-22
implement plotly_editor_traits

nicolaskruchten:
proposed solution to #258 (more robust interactions with children)

### DIFF
--- a/src/components/containers/Panel.js
+++ b/src/components/containers/Panel.js
@@ -38,28 +38,14 @@ class Panel extends Component {
     );
   }
 
-  isTraceAccordion() {
-    const {children} = this.props;
-    return (
-      children &&
-      !Array.isArray(children) &&
-      children.type.displayName &&
-      children.type.displayName.indexOf('TraceAccordion') >= 0
-    );
-  }
-
-  isAnnotationAccordion() {
-    const {children} = this.props;
-    return (
-      children &&
-      !Array.isArray(children) &&
-      children.type.displayName.indexOf('AnnotationAccordion') >= 0
-    );
-  }
-
   closeAllButLast(array) {
     const lastIndex = array.length - 1;
     return array.map((e, i) => i !== lastIndex);
+  }
+
+  firstChildWithCanAdd() {
+    const children = React.Children.map(this.props.children, c => c);
+    return children && children.filter(c => c.props.canAdd)[0];
   }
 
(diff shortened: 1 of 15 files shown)


nicolaskruchten:
This isn't particularly revolutionary or beautiful, but it does give us looser coupling than straight equality checks on `child.type` or indexOf checks on `child.type.displayName`...

bpostlethwaite:
Nope not a revolution but just the :ticket: ! Neatly captures a pattern. Any tests warranted? It would be nice to have something fail if one of the traits fails to get passed through during some future refactor.

nicolaskruchten:
This PR just leverages the existing tests basically.

src/components/containers/PanelHeader.js
```diff
@@ -74,13 +55,18 @@ class PanelHeader extends Component {
...
+PanelHeader.contextTypes = {
+  layout: PropTypes.object,
+  onUpdate: PropTypes.func,
+  updateContainer: PropTypes.func,
+};
+
 PanelHeader.propTypes = {
-  action: PropTypes.string,
+  addAction: PropTypes.object,
   allowCollapse: PropTypes.bool,
```
VeraZab: this is really nice! but should this 'action' stay more general, maybe it's not always going to be an 'addAction'

src/components/containers/Panel.js
```diff
@@ -100,40 +86,32 @@ class Panel extends Component {
...
-        return child;
-      });
+      );
 
       return (
         <div className={bem('panel')}>
           <PanelHeader
-            action={action}
+            addAction={addAction}
             allowCollapse={nbOfFolds > 1}
```
VeraZab: same here

src/components/containers/PanelHeader.js
nicolaskruchten: At the moment it's pretty well coupled to being `add` if we hardcode the `<PlusIcon>` like we do... Should I generalize the icon as well?

VeraZab: yeah, you're right, I didn't notice the icon, we should be generalizing that too then if going that route.

Well, ok, let's generalize when we need to generalize. I don't foresee us needing other actions in the panel header, other than add, anytime soon

VeraZab:
ok, well, I'm 💃  on this

PR #263, opened 2018-01-22
integrated development environment

nicolaskruchten:
Here is what I suggest for a quicker, `npm link`-free development setup :)


README.md
```diff
@@ -86,32 +86,10 @@ class App extends Component {
...
-
-```
-cd examples/simple # or any other application directory
 npm install
-rm -rf node_modules/react node_modules/react-dom node_modules/react-plotly.js-edit
-npm link react-plotly.js-edit
-npm start # keep this running
+npm start
+# start hacking
 ```
```
VeraZab: haha, well, yeah, a lot better
all the better that we do this before Jack releases all the marketing on this.

That way, if we're so lucky that others can't wait to contribute, well, there's an easier way to set all this up! Thanks!

nicolaskruchten: Would you use this for development going forward or is it still missing something that would keep you using the simple example with npm-link?

VeraZab: yes, let's use this going forward! thanks!

PR #264, opened 2018-01-22
Ohlc trace type

VeraZab:
ohlc attributes: https://github.com/plotly/plotly.js/blob/master/src/traces/ohlc/attributes.js

### DIFF
--- a/dev/index.js
+++ b/dev/index.js
@@ -1,6 +1,6 @@
 import React, {Component} from 'react';
 import ReactDOM from 'react-dom';
-import plotly from 'plotly.js/dist/plotly-basic';
+import plotly from 'plotly.js/dist/plotly';
 import createPlotComponent from 'react-plotly.js/factory';
 import PlotlyEditor from '../src';
 import '../src/styles/main.scss';
--- a/src/DefaultEditor.js
+++ b/src/DefaultEditor.js
@@ -48,6 +48,7 @@ class DefaultEditor extends Component {
               clearable={false}
               show
             />
+
             <DataSelector
               label="Labels"
               attr="labels"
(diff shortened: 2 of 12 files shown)


VeraZab:
ok @nicolaskruchten would you like to give this a review?
hopefully I left enough notes about why certain things were necessary and the plotly.js specificities for financial charts

if you have a moment to take a look @bpostlethwaite I wouldn't refuse either :)

src/DefaultEditor.js
```diff
@@ -164,6 +182,44 @@ class DefaultEditor extends Component {
...
+                label="Show in Legend"
+                attr="increasing.showlegend"
+                options={[
+                  {label: _('Show'), value: true},
+                  {label: _('Hide'), value: false},
+                ]}
+              />
+            </Section>
+
+            <Section name={_('Increasing Trace Styles')}>
```
nicolaskruchten: Should this be "Decreasing" instead of "Increasing" ?

VeraZab: oh, yes, 🙈

dev/index.js
```diff
@@ -1,6 +1,6 @@
 import React, {Component} from 'react';
 import ReactDOM from 'react-dom';
-import plotly from 'plotly.js/dist/plotly-basic';
+import plotly from 'plotly.js/dist/plotly';
```
nicolaskruchten: Hmmm this might cause an issue with webpack when uglifying... The test I'm running now is almost at 10 minutes and it's still not done :(

nicolaskruchten: `-basic` took 153s to complete with `webpack -p`

VeraZab:
I'll link this issue here for reference, about the difference of _fullInput vs _fullData: https://github.com/plotly/plotly.js/issues/2281

But, I know that for composed traced types, like financial charts, the true attributes of the trace are in _fullInput. However, _fullInput doesn't seem to contain everything that _fullData contains.

Well, we should be good for now I think :)

dev/index.js
VeraZab: but this is what I need to get all my other chart types..
if I don't have this, I only get the basic 5 chart types, I think it's ok that we add a fuller version into the dev env

nicolaskruchten: for the dev env ok but this won't work for something like the simple app: we'll need to load plotly.js the proper way i.e. not from `dist` :)

VeraZab: ah I see : ) ok, well, I'll merge this for now

PR #269, opened 2018-01-23
Add candlestick trace type

VeraZab:
https://github.com/plotly/plotly.js/blob/master/src/traces/candlestick/attributes.js


VeraZab:
@nicolaskruchten would you review please?

PR #271, opened 2018-01-24
Data Selector adjustments

VeraZab:
A few data selector adjustments that will be helpful for all trace types

### DIFF
--- a/src/DefaultEditor.js
+++ b/src/DefaultEditor.js
@@ -49,38 +49,23 @@ class DefaultEditor extends Component {
               show
             />
 
-            <DataSelector
-              label="Labels"
-              attr="labels"
-              clearable={false}
-              hasBlank
-            />
+            <DataSelector label="Labels" attr="labels" clearable hasBlank />
 
-            <DataSelector
-              label="Values"
-              attr="values"
-              clearable={false}
-              hasBlank
-            />
+            <DataSelector label="Values" attr="values" clearable hasBlank />
 
-            <DataSelector label="X" attr="x" clearable={false} hasBlank />
+            <DataSelector label="X" attr="x" clearable hasBlank />
 
-            <DataSelector label="Y" attr="y" clearable={false} hasBlank />
+            <DataSelector label="Y" attr="y" clearable hasBlank />
 
-            <DataSelector label="Z" attr="z" clearable={false} hasBlank />
+            <DataSelector label="Z" attr="z" clearable hasBlank />
 
-            <DataSelector label="Open" attr="open" clearable={false} hasBlank />
+            <DataSelector label="Open" attr="open" clearable hasBlank />
 
-            <DataSelector label="High" attr="high" clearable={false} hasBlank />
+            <DataSelector label="High" attr="high" clearable hasBlank />
 
-            <DataSelector label="Low" attr="low" clearable={false} hasBlank />
+            <DataSelector label="Low" attr="low" clearable hasBlank />
 
-            <DataSelector
-              label="Close"
-              attr="close"
-              clearable={false}
-              hasBlank
-            />
+            <DataSelector label="Close" attr="close" clearable hasBlank />
           </TraceAccordion>
         </Panel>
 
--- a/src/DefaultEditor.js
+++ b/src/DefaultEditor.js
@@ -42,30 +42,25 @@ class DefaultEditor extends Component {
       <PanelMenuWrapper>
         <Panel group="Graph" name="Create">
           <TraceAccordion canAdd>
-            <TraceSelector
-              label="Plot Type"
-              attr="type"
-              clearable={false}
-              show
-            />
+            <TraceSelector label="Plot Type" attr="type" show />
 
-            <DataSelector label="Labels" attr="labels" clearable hasBlank />
+            <DataSelector label="Labels" attr="labels" />
 
-            <DataSelector label="Values" attr="values" clearable hasBlank />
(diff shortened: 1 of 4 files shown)


VeraZab:
@bpostlethwaite would you 💃  on this one

src/components/fields/DataSelector.js
```diff
@@ -44,7 +44,16 @@ class DataSelector extends Component {
...
-      update[this.srcAttr] = value;
+      const srcAttr = this.srcAttr;
+      update[srcAttr] = value;
+
+      if (!value) {
+        /*
+         * we're deleting a data attribute, let's also clean it up from gd.data
+         * to prevent issues when switching plots (i.e. candlestick -> box plot)
+         */
+        update[srcAttr.replace('src', '')] = value;
```
bpostlethwaite: so value here is supposed to `undefined`? Should we make that more clear? What if `value=0` or `value=''` or other falsey values?

src/components/fields/DataSelector.js
```diff
@@ -44,7 +44,17 @@ class DataSelector extends Component {
...
+      const srcAttr = this.srcAttr;
+      update[srcAttr] = value;
+
+      if (!value) {
+        /*
+         * when !value, then we're deleting ${attr}src, let's also clan up the
+         * related src to prevent issues when switching plots (i.e. candlestick -> box plot)
+         * ex: clean up both xsrc and x from gd.data
+         */
+        update[srcAttr.replace('src', '')] = value;
```
nicolaskruchten: should this be just `update[this.props.attr] = null` ?

nicolaskruchten: or maybe merge this into the next if-block and always dereference it, even if it's null. that way if value is falsey then we get undefined back from dataSources.

src/DefaultEditor.js
```diff
@@ -49,38 +49,23 @@ class DefaultEditor extends Component {
               show
             />
 
-            <DataSelector
-              label="Labels"
-              attr="labels"
-              clearable={false}
-              hasBlank
-            />
+            <DataSelector label="Labels" attr="labels" clearable hasBlank />
```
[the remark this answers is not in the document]
VeraZab: I'm not sure about hasBlank, but to me, Data Selectors should all be clearable, because the user should be able to completely reset them, or not use some of the available ones.

The one that I think should not be clearable, is the trace selector dropdown. We always want for a trace type to be selected, and we have a default one, which is scatter, so there should be no situation where that dropdown appears, but no trace type is selected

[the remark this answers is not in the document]
nicolaskruchten: OK, so I vote we remove the `clearable` and `hasBlank` props.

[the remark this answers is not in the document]
bpostlethwaite: Yep I agree we should make that a default prop type then we can take out `clearable` all together (except `<TraceSelector clearable={false} />`)

src/components/fields/DataSelector.js
VeraZab: value gets set to null (from react-select) when you clear the dropdown (reset it completely)
and that is the situation that we're interested here, because when you clear a data field, you don't want gd.data to show x [ ], you'd prefer for gd.data to not have x at all, 

This is helpful for situations where you changed from one trace type that had x as a possible attribute to another trace that has that attribute, (candlestick -> box) but it's used in another way and so it can break the plot

VeraZab: I updated the comment

src/DefaultEditor.js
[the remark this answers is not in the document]
bpostlethwaite: Though TraceSelector could itself default to `clearable={false}` so that even that is unnecessary.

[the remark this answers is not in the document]
VeraZab: ok, hasBlank doesn't seem to be doing anything, I'll remove it

src/components/fields/DataSelector.js
```diff
@@ -44,7 +44,17 @@ class DataSelector extends Component {
     }
     const update = {};
     if (this.dataSrcExists) {
-      update[this.srcAttr] = value;
+      const srcAttr = this.srcAttr;
+      update[srcAttr] = value;
+
+      if (!value) {
+        /*
```
nicolaskruchten: I don't know if you saw my comments on this block, but I think it can just be folded into the next one... basically just remove this if and the next and replace them with just `update[this.props.attr] = this.dataSources[value]`.

This way, if `value` is falsey or otherwise not contained in the datasource, the `attr` is automatically cleared, and we don't need to do string manipulation on `srcAttr`.

src/components/fields/DataSelector.js
```diff
@@ -43,22 +43,10 @@ class DataSelector extends Component {
...
-        update[this.props.attr] = value;
-      }
-    }
 
-    if (Array.isArray(this.dataSources[value])) {
-      update[this.props.attr] = this.dataSources[value];
+    if (this.dataSrcExists) {
+      update[this.srcAttr] = value;
+      update[this.props.attr] = this.dataSources[value] || null;
     }
```
VeraZab: plotly.js needs a `null` to be sent for it to remove an attribute

VeraZab:
ok how is this one now @nicolaskruchten ?

nicolaskruchten:
💃

VeraZab:
yeah, that approve button has to be pressed, but it's ok, i can press dismiss to merge this time

PR #274, opened 2018-01-24
TraceTypeSection component

nicolaskruchten:
What do you guys think?


nicolaskruchten:
My thinking is that we wouldn't have to use this everywhere in Style > Traces, but only for pairs of sections which contain the same `attr`s in different contexts...

PR #275, opened 2018-01-25
break up default editor

nicolaskruchten:


src/index.js
```diff
@@ -96,6 +105,12 @@ export {
   AxesFold,
   Button,
   SingleSidebarItem,
+  GraphCreatePanel,
+  StyleLayoutPanel,
+  StyleAxesPanel,
+  StyleLegendPanel,
+  StyleNotesPanel,
+  StyleTracesPanel,
 };
```
VeraZab: Ah yeah, it's nice that they're exposed now so users can pick and chose the ones they'd like to use/replace

VeraZab: should we add a note to the readme?

nicolaskruchten: Not really necessary at this point, I don't think, it's mostly just easier for us to work with :)

PR #276, opened 2018-01-25
clean up smells in Section

nicolaskruchten:


nicolaskruchten:
The tests seem good enough to me to make this a safe refactor, and the line count is high just because I've crunched things down a bit with less-branchy conditionals, maps and fewer intermediate variables ;)

PR #283, opened 2018-01-25
Partial implementation of z-types

nicolaskruchten:
Big chunks of Heatmap, Contour and Surface are in, still need to add a bunch of stuff to Axes and track down some bugs.


VeraZab:
this looks good to me! 💃 
at the end of all trace types, we'll play around with the widgets look for those oddities we've spoken about, but I say this is good to merge for now.

PR #286, opened 2018-01-27
Timeseries trace added

VeraZab:
ex

### DIFF
--- a/src/lib/customTraceType.js
+++ b/src/lib/customTraceType.js
@@ -6,6 +6,8 @@ export function plotlyTraceToCustomTrace(trace) {
     )
   ) {
     return 'area';
+  } else if (trace.type === 'timeseries') {
+    return 'timeseries';
   } else if (
     trace.type === 'scatter' &&
     (trace.mode === 'lines' || trace.mode === 'lines+markers')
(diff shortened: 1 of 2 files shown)


src/lib/customTraceType.js
```diff
@@ -6,6 +6,8 @@ export function plotlyTraceToCustomTrace(trace) {
     )
   ) {
     return 'area';
+  } else if (trace.type === 'timeseries') {
+    return 'timeseries';
   } else if (
```
VeraZab: this is a hack, because even with autobinx and autobiny, I'm having trouble differentiating between line traces and timeseries, when I'm switching away from timeseries to line traces. 

Since autobinx and autobiny have been set when we switched to the timeseries trace, and are not erased anywhere, the if conditional recognizes a timeseries trace, even if the intention was to draw a line trace.

This works because when plotly.js doesn't recognize a trace type, it sets it to scatter. Maybe I could do the same hack with other custom scatter types. Seems like some confusions might happen when switching multiple times between different custom scatter types.

VeraZab:
@bpostlethwaite @nicolaskruchten ready for review

nicolaskruchten:
💃 Seems OK to me to merge at the moment but I'm not as familiar with the logic around timeseries/financial charts yet...

VeraZab:
I could merge this in now, and at the refactoring pass at the end, rediscuss if the timeseries hack is something that could be extended to the other custom scatter types, for consistency.

PR #288, opened 2018-01-29
Colorscale widget

VeraZab:
closed state
screen shot 2018-01-29 at 2 12 02 pm

open state
screen shot 2018-01-29 at 2 10 08 pm


src/styles/components/widgets/_colorscalepicker.scss
```diff
@@ -0,0 +1,17 @@
...
+  width: 5px;
+}
+
+.colorscale-selected {
+  display: none;
+}
+
+.rc-slider {
+  width: 200px;
+}
```
VeraZab: @aulneau could I have written any of this better?
it's to override the react-colorscales npm module styles..

src/components/fields/Colorscale.js
```diff
@@ -0,0 +1,43 @@
...
+    const {container, fullValue} = this.props;
+    const colorscale = fullValue.map(v => v[1]);
+    return container.marker && Array.isArray(container.marker.color) ? (
+      <Field {...this.props}>
+        <ColorscalePicker
+          selected={colorscale}
+          onColorscaleChange={this.onUpdate}
+        />
+      </Field>
+    ) : null;
```
nicolaskruchten: in general, for readability reasons, I prefer to have the shorter branch show up first, so `test ? null : ( long thing )`

VeraZab:
few notes: 
I've adjusted visibility of the color picker and colorscale picker. I make the color picker appear if marker.color is not an array, and if it's an array then the colorpicker appears. This may not work for all trace types though, so may need to be readjusted. 

Also, we may want to display both color picker and colorscale picker at the same time. But I'm not sure which plotly.js attributes I'd need to change to allow me to switch back to color by colorscale once I've already tried coloring all markers by a single color.

I'm thinking to maybe just letting this pr be about adding the colorpicker widget, and then adjusting color picker and color scale picker visibility at the end..

nicolaskruchten:
Makes sense for this PR to just be "adding widget". Maybe consider adding it to the documentation and the `custom` example so we have a copy-pastable example?

VeraZab:
also, I think some css needs to be adjusted in the npm module, sometimes it has way too many colors in its scales, i'll look into that.
screen shot 2018-01-29 at 2 20 21 pm

VeraZab:
yeah @nicolaskruchten I'm going to add this now to the readme and custom example : )

VeraZab:
ok I fixed added a pr to react-colorscales to limit to 9 the number of colors in the colorscale previewers:
https://github.com/plotly/react-colorscales/pull/2

It looks ok now:
screen shot 2018-01-29 at 6 43 32 pm

VeraZab:
@nicolaskruchten let me know if we can merge

nicolaskruchten:
Yup! 💃

PR #292, opened 2018-01-31
Wip geo traces

bpostlethwaite:
Not a whole lot going on this PR yet. The Create panel functionality has been added. I noticed that if you select a USA type region and then select a Projection plotly.js spit errors. I think projection should be hidden by fullLayout when a non-world region is selected? 

Workspace also has issues in this regard it seems. As usual I recommend fixing funky behaviour at the source (plotly.js) instead of papering over it here.


nicolaskruchten:
Thanks Ben! Go on vacation now, we'll look after this for you :)

nicolaskruchten:
:dancer:

nicolaskruchten:
:dancer:

PR #294, opened 2018-01-31
 enables nested Panel/Fold/Panel/Fold

nicolaskruchten:
This PR changes the relationship between Panels and Folds for expand/collapse behaviour: Folds must now be direct children of Panels for this connection to be made, and the connection is made via `props` instead of `context` so that we can nest Panel/Fold/Panel/Fold etc without the expand/collapse behaviour bleeding over into children. 

This entails that Accordions must supply their own Panels instead of being contained by them. 

Making this change allows us to clean up the `canAdd` vs `plotly_editor_traits.add_action` smell that the old setup forced us into :)

### DIFF
--- a/src/components/fields/TraceSelector.js
+++ b/src/components/fields/TraceSelector.js
@@ -101,7 +101,11 @@ class TraceSelector extends Component {
   }
 
   setTraceDefaults(container, fullContainer, updateContainer) {
-    if (!container.mode && fullContainer._fullInput.type === 'scatter') {
+    if (
+      container.uid &&
+      !container.mode &&
+      fullContainer._fullInput.type === 'scatter'
+    ) {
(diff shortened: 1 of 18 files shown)


src/components/fields/TraceSelector.js
```diff
@@ -101,7 +101,11 @@ class TraceSelector extends Component {
   }
 
   setTraceDefaults(container, fullContainer, updateContainer) {
-    if (!container.mode && fullContainer._fullInput.type === 'scatter') {
+    if (
+      container.uid &&
+      !container.mode &&
```
VeraZab: why is container.uid needed now?

VeraZab:
ok 💃  thanks for all the cleanups!

PR #297, opened 2018-02-01
Geo projections

VeraZab:
opened an issue with plotly.js here: https://github.com/plotly/plotly.js/issues/2320
cc @nicolaskruchten


src/components/fields/Dropdown.js
```diff
@@ -6,16 +6,31 @@ import {connectToContainer} from 'lib';
...
+      }
     }
 
     return (
       <Field {...this.props}>
         <DropdownWidget
           backgroundDark={this.props.backgroundDark}
-          options={this.props.options}
+          options={options}
           value={this.props.fullValue}
```
nicolaskruchten: if you're pulling stuff out of `props` you may as well reuse them here :)

VeraZab: oh forgot about this one : ) thanks!

nicolaskruchten: np. For what it's worth, I'm not a huge fan of giant `const {...} = this.props;` everywhere... I'm much happier just having `this.props.x` everywhere as it's less mental indirection when reading the code. Basically if I see `this.props.x` I *know* it's from props and hasn't been locally mutated. If I see just `x` I have to look up higher in the code to know where it came from... Is it `const`? Was it `let`ed and mutated 9 times? etc... :)

src/components/fields/Dropdown.js
```diff
@@ -11,11 +11,25 @@ export class UnconnectedDropdown extends Component {
       placeholder = this.props.fullValue;
     }
 
+    let options = this.props.options;
+
+    if (
+      this.props.attr === 'geo.projection.type' &&
```
nicolaskruchten: Is there any way we can do this outside of the `Dropdown`? I'd rather not bake in this super-specific attribute stuff. Can we create a `ProjectionsDropdown` instead?

src/default_panels/GraphCreatePanel.js
```diff
@@ -10,81 +10,122 @@ import {
...
+const GraphCreatePanel = ({localize: _}) => {
+  const geoProjectionOptions = {
+    usa: [
+      {
+        label: _('Albers USA'),
+        value: 'albers usa',
+      },
+    ],
+    other: [
+      {label: _('Equirectangular'), value: 'equirectangular'},
```
nicolaskruchten: Same here, can we push this all into a `ProjectionsDropdown` and a `GeoScopeDropdown` ?

nicolaskruchten: An additional advantage of this approach is that we could get rid of `LayoutSectionOverride` by just binding those two custom fields to the Layout.

src/components/index.js
```diff
@@ -5,32 +5,34 @@ import {
   AxesRange,
   AxesSelector,
   CanvasSize,
-  ContourNumeric,
   ColorPicker,
+  ColorscalePicker,
+  ContourNumeric,
```
nicolaskruchten: Which ESLint setting are you using that reorders imports like this? I'll turn it on myself so we get less churn like this where your editor fixes my mistakes in your commits :)

src/components/containers/derived.js
```diff
@@ -8,19 +8,12 @@ import {
...
-  supplyLayoutPlotProps,
 } from 'lib';
 
 const LayoutPanel = connectLayoutToPlot(Panel);
 
 const AxesFold = connectAxesToLayout(Fold);
 
-// Special purpose Section for usage when nested inside a Trace connected container.
-// supplyLayoutPlotProps overrides the trace context with a layout context.
-const LayoutSectionOverride = connectLayoutToPlot(Section, {
```
nicolaskruchten: yay!

src/components/fields/derived.js
```diff
@@ -234,3 +235,93 @@ function computeAxesRefOptions(axes) {
...
+        {label: _('Stereographic'), value: 'stereographic'},
+        {label: _('Mollweide'), value: 'mollweide'},
+        {label: _('Hammer'), value: 'hammer'},
+        {label: _('Transverse Mercator'), value: 'transverseMercator'},
+        {label: _('Winkel Tripel'), value: 'winkel3'},
+        {label: _('Aitoff'), value: 'aitoff'},
+        {label: _('Sinusoidal'), value: 'sinusoidal'},
+      ];
+
+      if (
```
nicolaskruchten: oh yeah, this is so much nicer than in `Dropdown` right? :)

VeraZab: right :)

PR #298, opened 2018-02-01
Fold hiding

nicolaskruchten:


src/components/fields/TraceSelector.js
```diff
@@ -154,6 +154,9 @@ TraceSelector.propTypes = {
 
 export default connectToContainer(localize(TraceSelector), {
   modifyPlotProps: (props, context) => {
+    if (!context.container) {
```
nicolaskruchten: more defensive coding :)

src/lib/unpackPlotProps.js
```diff
@@ -6,7 +6,7 @@ export default function unpackPlotProps(props, context) {
   const {container, getValObject, defaultContainer, updateContainer} = context;
 
   if (!props.attr) {
-    throw new Error('connectedToContainer components require an `attr` prop');
+    return {};
```
nicolaskruchten: this change required so that `<Section>` without `attr` wouldn't throw, and it's why we deleted the test above.

src/styles/components/containers/_fold.scss
```diff
@@ -159,6 +159,12 @@
       border-width: 1px 1px 1px 1px;
     }
   }
+
+  .panel {
```
nicolaskruchten: a bit gross but needed in order to style nested panels without breaking stuff like the color scale picker

src/styles/components/containers/_panel.scss
```diff
@@ -7,6 +7,7 @@
   border-right: var(--border-default);
   box-sizing: border-box;
   position: relative;
+  width: calc(var(--panel-width));
```
nicolaskruchten: restoring style I deleted earlier which was necessary for the color scale picker

src/components/fields/TraceSelector.js
VeraZab: that's what was the bug with timeseries?

src/default_panels/StyleColorbarsPanel.js
```diff
@@ -6,17 +6,17 @@ import {Radio, TextEditor, TraceAccordion, Fold, Panel} from '../components';
...
-    <Panel>
-      <Radio
-        attr="showscale"
-        options={[
-          {label: _('Show'), value: true},
-          {label: _('Hide'), value: false},
-        ]}
-      />
+  <TraceAccordion messageIfEmptyFold="Need a color for a colorbar!">
+    <Radio
```
VeraZab: but shouldn't this message be more individualized, what if it's not a color bar fold that is empty..

src/components/fields/TraceSelector.js
nicolaskruchten: no I don't think so, it was necessary for this PR to work

src/default_panels/StyleColorbarsPanel.js
nicolaskruchten: This is in the ColorBar panel only...

VeraZab: and it's needed in TraceAccordion? I thought TraceAccordion was just had a colorscale widget for when there's marker.colorscale..? there's a whole fold for color bars in traceAccordion?

src/lib/unpackPlotProps.js
VeraZab: yeah, ok, i've been needing unpackPlotProps to not require an attr as well, so ok

src/default_panels/StyleColorbarsPanel.js
nicolaskruchten: The TraceAccordion generates the panel and the folds so if we want the individual colorbar folds to have a message-if-empty this is where to pass it in. But look at the filename: this is a highly-inidividualized case :)

src/components/fields/TraceSelector.js
nicolaskruchten: this definitely does not fix the Timeseries issue, just checked, but it does address another crash pathway.

src/default_panels/StyleColorbarsPanel.js
VeraZab: oh sorry, didn't see the filename `StyleColorbarsPanel.js` 👍

VeraZab:
ok 💃

PR #301, opened 2018-02-05
Scattermapbox

VeraZab:
cc @nicolaskruchten


dev/index.js
```diff
@@ -96,6 +97,7 @@ class App extends Component {
               onInitialized={this.handlePlotUpdate.bind(this)}
               revision={this.state.plotRevision}
               style={{width: '100%', height: '100%', minHeight: '100vh'}}
+              config={{mapboxAccessToken: getAccessToken('MAPBOX')}}
```
nicolaskruchten: Ah, I figured this would happen under the hood, so the developer wouldn't need to do this, it would just get auto-loaded...

nicolaskruchten: If we're doing it this way, we can skip the `getAccessToken` part and just say they need to pass in a token and let them get it however they like.

nicolaskruchten: Also, I think you forgot to commit the file which defines `getAccessToken`, or at least I can't find its definition in the PR.

nicolaskruchten:
Looks great! 💃

PR #302, opened 2018-02-05
adding an error boundary in Panel

nicolaskruchten:
image

### DIFF
--- a/src/components/containers/Panel.js
+++ b/src/components/containers/Panel.js
@@ -1,17 +1,44 @@
 import PanelHeader from './PanelHeader';
+import PanelEmpty from './PanelEmpty';
 import PropTypes from 'prop-types';
 import React, {Component, cloneElement} from 'react';
 import update from 'immutability-helper';
-import {bem} from 'lib';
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
image

src/components/containers/Panel.js
```diff
@@ -1,19 +1,46 @@
...
       individualFoldStates: [],
+      hasError: false,
     };
     this.toggleFolds = this.toggleFolds.bind(this);
     this.toggleFold = this.toggleFold.bind(this);
   }
 
+  componentDidCatch() {
+    this.setState({hasError: true});
+  }
```
VeraZab: was there any more details we could have logged with console.error? componentDidCatch does take an error argument that maybe we could log?

nicolaskruchten: it doesn't log anything more than that screenshot I posted, just in a JSON object, so I didn't include it. It would/will be useful to log this kind of error to a production-monitoring system or something like that in the future. Like Errbit or Rollbar.

VeraZab:
💃

PR #303, opened 2018-02-05
Ternary

VeraZab:


VeraZab:
Hey @nicolaskruchten I think I'd like to handle axes adjustments separately, and merge this one in before. Axes are also broken on geo plots, so next pr is going to be fixing axes for geo and ternary.

nicolaskruchten:
works for me! 💃

PR #307, opened 2018-02-06
remove child.type checks

nicolaskruchten:
This completes the `plotly_editor_traits` refactoring, and plays nice with HMR. The `SidebarSection` thing breaks the API a tiny bit but really this is not the place to be adding classes, it should be done in the calling code.

### DIFF
--- a/src/components/containers/SingleSidebarItem.js
+++ b/src/components/containers/SingleSidebarItem.js
@@ -1,18 +1,10 @@
 import PropTypes from 'prop-types';
-import React, {cloneElement, Component} from 'react';
-import Button from '../widgets/Button';
+import React, {Component} from 'react';
 
 export default class SingleSidebarItem extends Component {
   render() {
-    let {children} = this.props;
-    children = React.Children.map(children, child => {
-      if (child.type === Button) {
-        return cloneElement(child, {className: 'button--menu'});
-      }
-      return child;
-    });
-    return children ? (
-      <div className="sidebar__item--single">{children}</div>
+    return this.props.children ? (
+      <div className="sidebar__item--single">{this.props.children}</div>
     ) : null;
   }
 }
--- a/src/styles/components/widgets/_button.scss
+++ b/src/styles/components/widgets/_button.scss
@@ -24,6 +24,7 @@ button {
 }
 
 .button {
+  $c: &;
   &__wrapper {
     display: flex;
     align-items: center;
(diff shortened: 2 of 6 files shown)


src/components/containers/SingleSidebarItem.js
```diff
@@ -1,18 +1,10 @@
...
-import React, {cloneElement, Component} from 'react';
-import Button from '../widgets/Button';
+import React, {Component} from 'react';
 
 export default class SingleSidebarItem extends Component {
   render() {
-    let {children} = this.props;
-    children = React.Children.map(children, child => {
-      if (child.type === Button) {
-        return cloneElement(child, {className: 'button--menu'});
```
VeraZab: but wait, aren't we losing some special styles on the buttons?

nicolaskruchten: Yes, but as I said in the header, this is the wrong place to do it... if you want special styles on your button, you should do it either in a CSS selector, or in the `<Button className={}>` but not magically here...

nicolaskruchten: We can't start adding special cases for every widget type that someone wants to jam into a sidebar :)

aulneau: We could fix this by using css rather than js to denote if its contained within the sidemenu, eg

```
.button{
  @at-root .sidemenu &{
    ... changes
  }
}
```

nicolaskruchten:
Are you ok with doing it in CSS? Or do you want another editor_trait?

src/components/containers/SingleSidebarItem.js
VeraZab: oh oups, was not seeing these..ok for the css

VeraZab:
yes ok for the css, didn't see the updated convo thread

src/components/containers/SingleSidebarItem.js
nicolaskruchten: @aulneau I'm actually not familiar with this `@at-root` directive... would it be possible for you to tack on a commit to this PR that does this please? :)

aulneau: sure thing!

nicolaskruchten: 🙇  normally I'm not one to skip on learning new stuff but I've got a few other things going on all of a sudden that people want my input on :)

aulneau: Okay! should be good to go @nicolaskruchten

src/styles/components/widgets/_button.scss
```diff
@@ -24,6 +24,7 @@ button {
 }
 
 .button {
+  $c: &;
   &__wrapper {
```
VeraZab: wow, thanks, what is this $c: ...

VeraZab:
thanks @aulneau @nicolaskruchten !

src/styles/components/widgets/_button.scss
aulneau: ah yeah I was going to use that

with some of my other projects that use scss i tend to write my styles where the name is only declared once, and then I attach it to a variable and can use it later like so:

carbon 32

Issue #238, opened 2018-01-13
Enhance chart picker dropdown


jackparmer:
The current chart picker dropdown won't scale to all of plotly's trace types when reintegrated with chart studio. I suggest making a design similar to the Feed and shipping it by default with this project:

![image](https://user-images.githubusercontent.com/1865834/34900314-ad2a29e2-f7b3-11e7-89fc-ffca955080bd.png)

For which the icons [are here](https://github.com/plotly/streambed/blob/master/shelly/webapp/static/webapp/fonts/new-plots/fonts/new-chart-icons.svg)

Developer requirements:
- [ ] Configurability of the chart names and which charts are displayed. Then we can eventually pass on this configurability to on-premises users who have corporate guidelines for how charts are named.
- [ ] Configurability of the what the columns are called, their order, and which chart go in which column.
- [ ] Bring-your-own chart picker components should still be technically possible for users who do not like this design, would prefer a simple dropdown. etc.

Related to: https://github.com/plotly/plotly.js/issues/2221

@fabric-8 is looking into design improvements.

jackparmer:
Added some of the new charts types on the horizon to this font file (mainly polar and 3d volume types). Thought this may be useful as either icon placeholders or the icons used in the final version.

http://plotly-icons.getforge.io/

[Plotly Icons.zip](https://github.com/plotly/react-plotly.js-editor/files/1630024/Plotly.Icons.zip)

aulneau:
@jackparmer nice! we've created a new repo for icons to be importable via `import {PlotTypeIcon} from 'plotly-icons';` which you can see here -> https://github.com/plotly/plotly-icons and see previews of them [here](https://plotly-icons.ineffable.digital). This is how icons are being displayed in the `react-plotly.js-editor`.

I went through all the icons I could find in `streambed` and added them, but I'll make a PR with any new icons from this zip.

fabric-8:
So here's the rebrush:
https://www.figma.com/file/fZWLt6JFr2lKxJNcP3CbCX6m/Streambed?node-id=223%3A1
<img width="1200" alt="modal_chart_picker 2x" src="https://user-images.githubusercontent.com/4112698/35249049-c5b2b46e-ffd0-11e7-8e83-23b2b521b3da.png">

Idea is to go for a modal instead of "dropdown" as it's easier to have additional interactions in there (show example / search / etc. icons when hovering a chart) and allows for more real estate.

[...]

PR #311, opened 2018-02-07
[WIP] Trace picker updates

aulneau:
Okay!

So this is _not_ quite ready yet. I've built out the views and I built out a little modal provider that we can use with other modals if we want to keep it. 

chart-editor

General structure of the updates:

I've added two new props to `PlotlyEditor`:
[`advancedTraceTypeSelector`](https://github.com/plotly/react-plotly.js-editor/blob/trace-picker-updates/src/PlotlyEditor.js#L212) which enables everything, and [`traceSelectorConfig`](https://github.com/plotly/react-plotly.js-editor/blob/trace-picker-updates/src/PlotlyEditor.js#L205) which passes our default configs to the enhanced trace type selector. The default config values are coming from some constants I ported over from `streambed` which you can see [here](https://github.com/plotly/react-plotly.js-editor/blob/trace-picker-updates/src/lib/constants.js#L51). Let me know if this data structure should change.

I have created a [`ModalProvider.js`](https://github.com/plotly/react-plotly.js-editor/blob/trace-picker-updates/src/components/containers/ModalProvider.js) which passes two functions through context to anywhere in the application: `openModal(component)` and `closeModal()`. We call `openModal` and pass it a component -> [see here](https://github.com/plotly/react-plotly.js-editor/blob/trace-picker-updates/src/components/fields/TraceSelector.js#L155).

I have also created a [`Modal.js`](https://github.com/plotly/react-plotly.js-editor/blob/trace-picker-updates/src/components/containers/Modal.js) component that we can pass params to. Currently it accepts a `title` param for the header, and then we can pass it content via `children` wrapped in `<ModalContent>...</ModalContent>` I might just bake this in rather than using a sep wrapper component for this. The `Modal.js` takes care of animating in/out.

Where I need help // cc @VeraZab @nicolaskruchten 
I am not exactly sure the best way to hook up the [`TraceTypeSelector.js`](https://github.com/plotly/react-plotly.js-editor/blob/trace-picker-updates/src/components/widgets/TraceTypeSelector.js) to be able to affect the larger application state (change chart types, only allow X type charts if Y type chart is already selected ([see here](https://github.com/plotly/react-plotly.js-editor/issues/238#issuecomment-363446538)), etc. Currently I am passing the [same props](https://github.com/plotly/react-plotly.js-editor/blob/trace-picker-updates/src/components/fields/TraceSelector.js#L155) that we pass to the drop down to the `TraceTypeSelector.js`, but I will need some guidance moving forward.

Things I still need to do:
- [x] Hook up the enhanced picker to actually change the trace type
- [ ] Convert the new chart type images to react svg components (and add the last 2 from @fabric-8)
- [ ] Provide a way for people to pass in custom images along with their custom data
- [ ] Possible default image/icon for a chart type that is passed but does not have a plotly image / provided image
- [x] Allow [item actions](https://github.com/plotly/react-plotly.js-editor/blob/trace-picker-updates/src/components/widgets/TraceTypeSelector.js#L45) to be customizable

Will resolve: #238


VeraZab:
> Convert the new chart type images to react svg components
do you mean by this: put these icons into the `plotly-icons` repo?

aulneau:
@VeraZab possibly -- these svgs are a lot more complex (various strokes, different colors) -- but it might make sense to add them. I'll have to either manually make them into components or adjust our generating scripts for them.

VeraZab:
To make the Modal change the plot, you have the updateContainer prop, you can pass it the value of the chart that the user clicks on. So for ex, for a timeseries chart, we have to make sure the timeseries value is passed on. It's a custom type. Careful with the custom types. We have that function `plotlyTraceToCustomTrace` to make those conversions, it takes container as an argument (gd.data).

VeraZab:
> but it might make sense to add them there.

yup, I think it does make sense to add them there

nicolaskruchten:
A few comments:

1. Looks great!
2. regarding the 2d/3d chart exclusion, basically in this component there needs to be some check of what's currently in the figure, and then some decision about which chart types to grey out. We can implement this after the basic version is done in a separate PR, but I don't see this as "affects the larger application state" so much as "conditionally renders based on state" if you see what I mean...
3. (Vera beat me to it)

src/DefaultEditor.js
```diff
@@ -13,15 +13,17 @@ import {
...
-    <GraphCreatePanel group={_('Graph')} name={_('Create')} />
-    <StyleTracesPanel group={_('Style')} name={_('Traces')} />
-    <StyleLayoutPanel group={_('Style')} name={_('Layout')} />
-    <StyleNotesPanel group={_('Style')} name={_('Notes')} />
-    <StyleAxesPanel group={_('Style')} name={_('Axes')} />
-    <StyleLegendPanel group={_('Style')} name={_('Legend')} />
-    <StyleColorbarsPanel group={_('Style')} name={_('Color Bars')} />
-  </PanelMenuWrapper>
+  <Fragment>
+    <PanelMenuWrapper>
```
VeraZab: is Fragment needed there?
If I understand correctly, Fragment's used to group elements in place of wrapping them in a div, but I don't see why it'd be useful here..?

aulneau: Whoops that's left over from when I was exploring ways of adding the modal :)

src/PlotlyEditor.js
```diff
@@ -1,11 +1,13 @@
...
 import PropTypes from 'prop-types';
-import React, {Component} from 'react';
+import React, {Component, Fragment} from 'react';
 import {bem} from './lib';
 import {noShame, maybeClearAxisTypes} from './shame';
 import {EDITOR_ACTIONS} from './lib/constants';
 import isNumeric from 'fast-isnumeric';
 import nestedProperty from 'plotly.js/src/lib/nested_property';
+import {CATEGORY_LAYOUT, TRACE_TYPES} from 'lib/constants';
+import {ModalProvider} from 'components/containers';
```
VeraZab: we have a modal system in webapp, and I think that with a modal system being present now in the editor, this just means that both modal systems will have to coexist. Because the biggest thing in webapp with Modals was that we wanted to have Modal action chains. 

One modal appears, then based on user's action, another one displays. But I think it's ok not to worry about it too much here. I think that what will most likely happen is that there's 2 modal systems that will exist.

aulneau: Yeah I was curious about that, I figured that in this version of the editor we didn't need chained modals, but I think we could merge it into one when the time comes :)

src/components/fields/TraceSelector.js
```diff
@@ -144,12 +145,28 @@ class TraceSelector extends Component {
...
+    const {advancedTraceTypeSelector} = this.context;
+    if (advancedTraceTypeSelector) {
+      return (
+        <div
+          className="trace-type-select-dropdown__wrapper"
+          onClick={() =>
+            this.context.openModal(<TraceTypeSelector {...props} />)
+          }
+        >
+          <UnconnectedDropdown {...props} />
```
VeraZab: Just from a UI and visual standpoint, maybe it should look more like a button here, instead of an UnconnectedDropdown?

nicolaskruchten:
re how to wire things in to the bigger context, take a look at how the other widget/field pairs work.

basically your `TraceTypeSelector` is an unconnected widget, and the `TraceSelector` is a field which connects it. So the field will pass you a prop of your choice bound to its `props.updatePlot` which you need to call whenever the user makes a choice.

src/components/widgets/TraceTypeSelector.js
```diff
@@ -0,0 +1,129 @@
...
+          target="_blank"
+        >
+          <SearchIcon />
+        </a>
+        <div
+          className="trace-item__actions__item"
+          aria-label="View tutorials on this chart type."
+          data-microtip-position={`top${position}`}
+          role="tooltip"
+        >
```
VeraZab: ok, you'd add the tutorial link later? just a reminder :) it should be an `a` tag?
and actually, I think this text _('View tutorials on this chart type') has to be localized.

There's a few places in here where text needs to be localized : )

src/components/widgets/TraceTypeSelector.js
```diff
@@ -0,0 +1,129 @@
...
+          <ThumnailViewIcon />
+        </div>
+        <div
+          className="trace-item__actions__item"
+          aria-label="See a basic example."
+          data-microtip-position={`bottom${position}`}
```
VeraZab: localized

src/components/widgets/TraceTypeSelector.js
aulneau: Yeah, I think I am going to abstract these out anyway to they are a part of the config that a user can pass through... enable them, change them, etc

src/components/fields/TraceSelector.js
aulneau: Cool -- I will build out a button similar to what is in the workspace currently :)

src/lib/constants.js
```diff
@@ -46,3 +46,241 @@ export const RETURN_KEY = 'Enter';
...
+// Layout specification for TraceTypeSelector.js
+export const CHART_CATEGORY = {
+  BUSINESS: 'BUSINESS',
+  SCIENCE: 'SCIENCE',
+  CHARTS_3D: 'CHARTS_3D',
+  FINANCIAL: 'FINANCIAL',
+  STATISTICS: 'STATISTICS',
+  MAPS: 'MAPS',
+};
+
```
VeraZab: Could these structures be simpler, like:
```
export const CHART_CLASS = {
 GEOGRAPHIC: ['atlas', 'choropleth', 'scattermapbox']
}
```

CHART_CATEGORY is needed for the picker and CHART_CLASS for compatibility and disabling and enabling different plots. So I see why both are needed. But I don't see why we should repeat that whole object that lists all chart types..

src/lib/constants.js
```diff
@@ -46,3 +46,241 @@ export const RETURN_KEY = 'Enter';
...
+export const CATEGORY_LAYOUT = [
+  {category: CHART_CATEGORY.BUSINESS, label: 'Business'},
```
VeraZab: These Category layouts need to be localized

src/styles/components/containers/_modal.scss
```diff
@@ -0,0 +1,99 @@
+.modal {
+  $c: &;
+  position: fixed;
```
VeraZab: is this a forgotten variable?

src/lib/constants.js
```diff
@@ -46,3 +46,241 @@ export const RETURN_KEY = 'Enter';
...
+  {category: CHART_CATEGORY.SCIENCE, label: 'Science'},
+  {category: CHART_CATEGORY.CHARTS_3D, label: '3D charts'},
+  {category: CHART_CATEGORY.FINANCIAL, label: 'Finance'},
+  {category: CHART_CATEGORY.STATISTICS, label: 'Statistics'},
+  {category: CHART_CATEGORY.MAPS, label: 'Maps'},
+];
+
+export const TRACE_TYPES = {
+  scatter: {
+    meta: {
```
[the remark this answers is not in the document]
aulneau: Much of these consts were pulled from `streambed` I was not sure how much of it should stay / how much of it should go. 

What do you think about moving https://github.com/plotly/react-plotly.js-editor/blob/master/src/components/fields/TraceSelector.js#L21 into the `constants.js` file?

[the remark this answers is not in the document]
VeraZab: We don't have to use what streambed has : ) we just need the logic (basically which chart types go into which categories). Plus our new requirement now is that users can make that configurable. So the best data structure is the one that's simplest to use in an API.

About https://github.com/plotly/react-plotly.js-editor/blob/master/src/components/fields/TraceSelector.js#L21 I think they're here because trace names need localization, and that localize function must wrap a component to make the localize prop available to it.

We should probably be making localize available at the very top level of our app, localizing component per component is a bit rough and unecesary..

I don't mind moving all trace types to the constants folder though..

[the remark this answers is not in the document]
nicolaskruchten: there should definitely be one canonical mapping from trace type to localized trace name, so that if I have access to `"scatter"` and `_` I should be able to get at the localized name. Makes sense to move that into `constants` and pack in the default categorization as well IMO.

VeraZab:
hey @aulneau!
we've discussed this a little bit here: https://github.com/plotly/react-plotly.js-editor/issues/260#issuecomment-364287520

And finally decided not to put these as separate trace types into the chart picker: 'error bars', 'animations', 'timeseries'. Would you remove them from the picker, please.

aulneau:
Okay I've started going through and creating lists based off of the icons that [Fab has made](https://github.com/plotly/react-plotly.js-editor/tree/efe456125811f5a48eeba8671422651376f25cc1/dev/_temp), and the [list I pulled from streambed](https://github.com/plotly/react-plotly.js-editor/blob/efe456125811f5a48eeba8671422651376f25cc1/src/lib/constants.js#L83), and the list contained [here in this editor](https://github.com/plotly/react-plotly.js-editor/blob/master/src/components/fields/TraceSelector.js#L21) and this is what I've got so far:

Trace types with category and icon -> [here](https://gist.github.com/aulneau/f0c51b07e1c85767a0d65c0347fd74f8)

Trace types with icon but _no category_ -> [here](https://gist.github.com/aulneau/5c2db04839e5a49513127744113e0e31)

Trace types with no icons and no category -> [here](https://gist.github.com/aulneau/82b22244d0059bff8c41d1c9d212ed9b)

Also @VeraZab and @nicolaskruchten do we want to couple the category type in with the trace type, eg:

**A**
```
    {
        value: 'scatter',
        label: _('Scatter'),
        category:CHART_CATEGORY.BUSINESS,
    },
```
or do we want to keep trace types clean and do something like:

**B**

```
    {
        value: 'scatter',
        label: _('Scatter'),
    },
---
const categories = [
  {
    value: 'statistics',
    label: _('Statistics'),
    traces: ['scatter', 'line', ...more],
  },
  ...
];
```
I am leaning more towards A.

nicolaskruchten:
I like A, it's DRYer

aulneau:
Okay -- I have added all the trace types that have icons into categories.

cc @VeraZab @nicolaskruchten @jackparmer for any suggestions on moving trace types to a more appropriate category:

screen shot 2018-02-09 at 10 46 32 am

nicolaskruchten:
Hmmm. How attached are we to our existing categorization scheme, @jackparmer ? It seems a bit limiting at this point if we're adding all the other trace types... 

Maybe we want to use a more function-oriented approach like this? https://datavizcatalogue.com/search.html or https://visual.ly/community/infographic/how/graphic-continuum

There are bunch of other organization types here https://github.com/widged/data-for-good/wiki/Visualisation-::-Choosing-a-chart

jackparmer:
> Hmmm. How attached are we to our existing categorization scheme

Not attached to it at all. I'm sure y'all can come up with something much better.

nicolaskruchten:
Here's a slightly different typology which includes all 3 carpet types and might fit into a 4-row/8-column grid.

* Simple (8, first row)
  * Scatter
  * Bar
  * Line
  * Area
  * Heatmap
  * Contour
  * Table
  * Pie
* 3D (5, second row)
  * 3D scatter
  * 3D line
  * 3D surface
  * 3D mesh
  * Pointcloud
* Maps (3, second row)
  * Choropleth
  * Satellite
  * Atlas
* Distributions (5, third row)
  * box
  * violin
  * histogram
  * 2d histogram
  * 2d contour histogram
* Finance (2, fourth row)
  * OHLC
  * Candlestick
* Specialized (7, fourth row)
  * Ternary
  * Polar
  * Parallel Coordinates
  * Sankey
  * Carpet
  * Scatter Carpet
  * Carpet Contour

jackparmer:
I like it! Only thing is `pointcloud` isn't 3d. I think we should have a 7th column called "2d WebGL" that has `scattergl`, `pointcloud`, `scatterpolargl`, and `heatmapgl`. This will be good for us - it's too hard to play around with the WebGL trace types currently.

nicolaskruchten:
I was thinking that the GL types would be like a checkbox or something rather than a first-class citizen in this trace-picker. The user looking at this might not want to make the "underlying technology" choice at this point...?

What's pointcloud if not 3-d?

jackparmer:
+1 Checkbox/switch sounds great to me. Maybe on the top-right here? One side of the switch says D3/SVG and the other side says WebGL? 

image

> What's pointcloud if not 3-d?

This magical trace type that @monfera wrote for uber refresh rates with lots of 2d scatter data:

https://codepen.io/monfera/pen/BLjJVZ
https://plot.ly/javascript/pointcloud/

If people want to plot a stupid # of x-y points interactively, we point them to `pointcloud`

nicolaskruchten:
One downside to the SVG vs WebGL thing at the trace-picker level is that then 3D would need to go under WebGL, right?

jackparmer:
I would label it "2d WebGL" and keep 3d charts under both views

nicolaskruchten:
OK so my initial thinking about the checkbox was outside of this modal. If we want to have the GL versions on this modal, I would go with a separate section called "WebGL accelerated" or something with those trace types in there.

jackparmer:
👍  "WebGL accelerated" column. I don't have a strong preference about the WebGL display as long as they are easy to access like the other trace types.

nicolaskruchten:
@aulneau do you have enough input to move forward with this PR? We will probably want to shuffle the categories again later in the future, so we probably don't need to nail it all down right now/this PR should leave things easy to change in that department :)

aulneau:
@nicolaskruchten yep! going to add some final touches and it should be ready for another review this afternoon :)

aulneau:
Okay, I've added the new categories and placed the trace types in them as we have decided (let me know if any should be moved around).

I think something we should discuss is if we want to have the amounts in each column so varied:

screen shot 2018-02-12 at 10 50 08 am

To try and prevent the columns from being so long, I've made a version that if there are more than 4 items in a column, that column will break into 2 sub-columns:

screen shot 2018-02-12 at 11 05 52 am

Thoughts? cc @VeraZab @jackparmer @nicolaskruchten

nicolaskruchten:
Oh my original suggestion had been different in two ways:

1. rows as opposed to columns
2. allow multiple groups per row, so that e.g. Maps and Distributions could sit on one row to make things less ragged

fabric-8:
@aulneau Yes that makes def. more sense given the uneven amount of items – Think the bar below the titles is not necessary though, but vertical bar helps to keep things seperate

Rows vs. Columns:
+ Row More flexible/robust in regards of layout, also better if trace types are being added or removed
- "Other" plot cats might get pushed out of the view, so user might have to scroll to see all cats

WebGL:
From a user perspective, how is WebGL different from 3d? Isn't that just a technical diff a user potentially doesn't really care about when choosing his trace type?

nicolaskruchten:
(porting some of my comments from private chat with @aulneau)

we could do something like a grid/block system:

image uploaded from ios

which would degrade reasonably well on narrow screens.

Also, keep in mind that this selector needs to be a bit dynamic: not every instantiation of this editor will want to have the same chart types/categorizations. So a dynamic type-to-category and category-to-layout specification would be good, even if we put a lot of though into the default one. Also we are likely to add chart types over time, so even if we land on something really balanced right now, it might break the minute we add a new type ;P

nicolaskruchten:
(also, +1 on merging WebGL and 3D)

nicolaskruchten:
Here are the potentially-upcoming chart types: https://github.com/plotly/plotly.js/issues/2221

fabric-8:
@nicolaskruchten @aulneau Def. like this approach, seems like the best thing to do 👍

jackparmer:
LGTM. Merging WebGL + 3d sounds good too, I don't have a strong opinion.
Keep in mind there will be 4 more 3d types by end of month (cone, volume,
isosurface, and streamtube)

aulneau:
Okay -- I have implemented the layout updates. Hurray for css grid! (will have to create a fallback flex version)

screen shot 2018-02-14 at 2 16 53 pm

The only things left I have are:

- [ ] get icons converted to react components (slightly hung up because these are more complex and include css in the svgs)
- [ ] abstract out icon to be replaced if the user wants to supply their own
- [ ] update the select to be a button when this trace picker is enabled

nicolaskruchten:
Oh yeah, I really like this new layout! :D

VeraZab:
looks sooo good @aulneau !!!

jackparmer:
He's a wizard! 🔮

fabric-8:
Very nice @aulneau! Also seems like we need a few more icons? :)

aulneau:
@fabric-8 I am currently using the same icons for `Scatter GL`, `Scatter Polar GL`, and `Heatmap GL`, but we don't have an icon to use for `Point Cloud`.

I've updated some of the icon names in `plotly-icons` to reflect all the trace types we have (I've also added a flag so we can render `simple` icons versus the new ones if we ever need to).

screen shot 2018-02-19 at 9 59 30 am

I've currently disabled the "Actions" functionality as I don't have a list of all the tutorials we want to link to for each trace type, and I am not sure how to implement the functionality we have in chart studio where when you click on `See a basic example.` it will load up that kind of trace type with data. Probs best to do the actions bit in a follow up PR so this one isn't delayed forever.

Fab is in the process of re-exporting out our SVG's so that there are no styles embedded in `style` tags, which has been preventing me from converting them to react components. Once those are good to go, this PR should be able to be merged (barring any new changes/fixes recommended via @VeraZab or @nicolaskruchten :) )

kapture 2018-02-19 at 10 09 29

nicolaskruchten:
Are we going to add all three carpet trace types (`carpet`, `scattercarpet`, `contourcarpet`) to this selector? Do we need separate icons for each?

nicolaskruchten:
Come to think of it, if Finance gets a section with 2 types and Maps with 3, might be worth adding a Carpet section with 3? That would unbalance the layout a bit though :(

aulneau:
Yeah -- I do think if we keep adding more and more sections, the general layout breaks down a bit. Also I think there might be an upper limit where this grid layout still makes sense vs a list view.

jackparmer:
The carpets are a bit of fringe trace type. Not huge use cases outsideof aerospace as far as I know.

One idea for carpets that I had is a "carousel" where you can click a left/right pointer to get a different "variety" of carpet plot, but this carousel would take up the same real estate as a normal icon. We could do the same thing for the 2d histograms.

nicolaskruchten:
Maybe instead of a carousel we could just have a fly-out menu appear when you click on certain trace types, to choose a specialization/variant. That would definitely save some space and work well with the WebGL-vs-SVG thing. We could group OHLC and candlestick into a single "Finance" icon. Box and Violin group nicely also etc.

VeraZab:
> Also I think there might be an upper limit where this grid layout still makes sense vs a list view.

Yeah @aulneau that would make sense to me

src/components/containers/Modal.js
```diff
@@ -0,0 +1,73 @@
...
+);
+
+const ModalContent = ({children}) => (
+  <div className="modal__content">{children}</div>
+);
+
+class Modal extends Component {
+  constructor(props) {
+    super(props);
+  }
```
VeraZab: we don't need the constructor, right : )

VeraZab:
so I've just commented out a few traces, because they're not going to go into the editor just now:
('scatterpolargl', 'pointcloud', 'scatterpolar', 'carpet', 'sankey', 'parcoords', 'violin')

and this is what the layout currently looks like without them:
screen shot 2018-02-20 at 10 01 26 am

Could we readjust a little

src/components/containers/ModalProvider.js
```diff
@@ -0,0 +1,95 @@
...
+    };
+  }
+
+  componentDidUpdate() {
+    const body = document.body;
+    const {open} = this.state;
+
+    // Toggle scroll on document body if modal is open
+    const hasClass = body.classList.contains('no-scroll');
+
```
VeraZab: you know, I'm not sure how much is possible, but could we try reducing margins in the modal a little so that there's no need to scroll by default?
It's a little sad to have to scroll everytime I pick a trace type..

src/components/fields/TraceSelector.js
```diff
@@ -170,4 +128,47 @@ TraceSelector.propTypes = {
...
 
-export default connectToContainer(localize(TraceSelector));
+export default connectToContainer(localize(TraceSelector), {
+  modifyPlotProps: (props, context) => {
+    if (!context.container) {
+      return;
+    }
+    if (
+      plotlyTraceToCustomTrace(context.container) === 'timeseries' &&
+      (!context.layout.xaxis ||
```
VeraZab: oh.. we removed this code.. maybe time for a little rebase?..

src/components/widgets/TraceTypeSelector.js
```diff
@@ -0,0 +1,165 @@
...
+    label: `View tutorials on this chart type.`,
+    href: `#`,
+    icon: <ThumnailViewIcon />,
+  },
+  {
+    label: `See a basic example.`,
+    href: `#`,
+    icon: <GraphIcon />,
+  },
+];
```
VeraZab: yeah, it's ok, we can add these in another pr : )

src/lib/computeTraceOptionsFromSchema.js
```diff
@@ -0,0 +1,150 @@
+function computeTraceOptionsFromSchema(schema, _, context) {
+  // Filter out Polar "area" type as it is fairly broken and we want to present
+  // scatter with fill as an "area" chart type for convenience.
+  const traceTypes = Object.keys(schema.traces).filter(
```
VeraZab: so, when is this code activated now?
When `advancedTraceTypeSelector` is not provided as a prop to the PlotlyEditor Component?
I think the user should still be able to configure the plot types even if they don't choose the `advancedTraceTypeSelector`.

So this should be adjusted and only read from schema if a trace config was not given.
And these should be the excluded types for now: 'area', 'violin', 'pointcloud', 'parcoords', 'sankey', 'carpet', 'scattercarpet', 'contourcarpet', 'scatterpolar'

src/lib/customTraceType.js
```diff
@@ -25,6 +25,8 @@ export function traceTypeToPlotlyInitFigure(traceType) {
       return {type: 'scatter', mode: 'markers', fill: 'none'};
     case 'area':
       return {type: 'scatter', fill: 'tozeroy'};
+    case 'cartesianArea':
+      return {type: 'scatter', fill: 'tozeroy'};
```
VeraZab: why was cartesianArea added?, 
and btw, these:
https://github.com/plotly/react-plotly.js-editor/blob/caf87d269d8bc270054bae0fc99b66bed464f64f/src/lib/customTraceType.js#L8

have to match these:
https://github.com/plotly/react-plotly.js-editor/blob/caf87d269d8bc270054bae0fc99b66bed464f64f/src/lib/customTraceType.js#L26

So area's already in there, we don't need 'cartesianArea'

src/lib/traceTypes.js
```diff
@@ -0,0 +1,207 @@
...
+    label: _('Scatter Polar GL'),
+    category: chartCategory(_).WEB_GL,
+  },
+  {
+    value: 'heatmapgl',
+    icon: 'heatmap',
+    label: _('Heatmap GL'),
+    category: chartCategory(_).WEB_GL,
+  },
+];
```
VeraZab: great : ) thanks!!

VeraZab:
I think I'd vote here to make these adjustments, and merge this pr in. We have all the icons for the currently integrated traces in the editor. We can make smaller prs to integrate the rest, and make layout improvements as we go.

src/lib/customTraceType.js
aulneau: Will remove :) it was held over from when I ported some stuff over from `streambed`

dev/App.js
```diff
@@ -83,6 +83,7 @@ class App extends Component {
             dataSources={dataSources}
             dataSourceOptions={dataSourceOptions}
             plotly={plotly}
+            advancedTraceTypeSelector
```
nicolaskruchten: Do we need this flag? this is basically going to be the default config right?

src/components/containers/ModalProvider.js
nicolaskruchten: +1 on faster fades, both in and out please!

dev/App.js
nicolaskruchten: ignore this comment, i have no objection here :)

VeraZab:
@aulneau checked out your branch and played around with it, looks good! let's merge!
and then when all icons are in plotly-icons we can make those changes in another pr. thanks!! 🎉

VeraZab:
ok, I think I resolved all rebase issues here, going to merge and if anything was missed then we'll see pretty soon :)

PR #160, opened 2017-12-06
Group style panels

VeraZab:

Issue #290, opened 2018-01-30
Fit'n'finish checklist


nicolaskruchten:
- [x] line colouring on fields for contour trace types
- [x] axis position field not hiding when axis is hidden
- [x] axis position field should be sensitive to which axis it is and show bot/top for x and left/right for y
- [x] default axis selector state for 3d 
- [x] exponents field not hidden when labels hidden
- [x] range slider fold not working
- [x] AxesFolds should autohide when empty
- [x] Layout panel in Axes missing 'anchor' field
- [x] step offset and step size should hide in Ticks when number of markers is custom
- [x] prefix/suffix fields should turn into text boxes when custom is selected
- [x] tickformat
- [x] missing flip vertical/horizontal controls in heatmap trace type
- [x] validate axes for ternary and pie
- [x] gaps in data selector appears the same for lines and gradients in contour
- [x] zmin/zmax don't hide when zauto is true
- [x] points color interferes with points colorscale and doesn't work with showscale
- [x] annotations: preconfigured annotation types
- [x] group style traces #160
- [x] empty menupanels shouldn't render a gear (i.e. tickprefixmode when no prefix present)
- [x] colorbar.tickfont.color not appearing, colorbar.tickprefix not disappearing
- [x] colorbar bgcolor doesn't work (or in workspace)
- [x] color picker at bottom of fold in panel in fold is clipped
- [x] `<br>` shouldn't show in rich text editor when empty (basically we should trim/strip the input)
- [x] hover projections/spikelines don't make sense on heatmaps (or aren't renderered)
- [x] step size on "fraction of plot" numerics should be 0.05 but sometimes they switch to pixel-mode due to another flag :( so we might need a compound control (e.g. in color-bar and shape and notes positioning)
- [x] clicking on checkbox label should select it
- [x] colorbars panel is empty for 3d scatter
- [x] what if there's 2 colorbars for one trace, like 3d scatters marker.color & line.color
- [x] mesh 3d: Contour on Hover, if option is at hide, all the contour details should now show (report plotly.js)
- [x] ensure that trace group names are localized
- [x] get proper default colors for symbol selector in multi-value context
- [x] remove exclusion of financial traces from grouped trace style tab
- [x] AxesSelector should select first non-All option by default
- [x] Shape default fill colour has alpha=0 ... odd to control opacity twice ... had to force defaults but it's weird to have to do this and for fill to appear on line shapes
- [x] Image defaults
[...]

PR #312, opened 2018-02-08
Grouped Trace tabs

nicolaskruchten:
This PR replaces #160. Multi-value UX similar to Axes panel yet to come, in another PR :)

### DIFF
--- a/src/components/containers/Panel.js
+++ b/src/components/containers/Panel.js
@@ -87,8 +87,6 @@ class Panel extends Component {
   render() {
     const {individualFoldStates, hasError} = this.state;
 
-    const {tabs} = this.props;
-
     if (hasError) {
       return <PanelError />;
     }
@@ -108,7 +106,9 @@ class Panel extends Component {
     );
 
     return (
-      <div className={`panel${tabs ? ' panel--with-tabs' : ''}`}>
+      <div
+        className={`panel${this.props.noPadding ? ' panel--no-padding' : ''}`}
+      >
         <PanelHeader
           addAction={this.props.addAction}
           allowCollapse={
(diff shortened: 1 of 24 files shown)


nicolaskruchten:
Oh I should mention that trace type names in a group context are not localized (or even properly capitalized!) yet but that will come once #311 is merged and I'll be able to leverage those localized constants from somewhere.

nicolaskruchten:
I've added an item to #290 to capture this.

src/components/containers/Panel.js
```diff
@@ -106,7 +108,7 @@ class Panel extends Component {
     );
 
     return (
-      <div className={bem('panel')}>
+      <div className={`panel${tabs ? ' panel--with-tabs' : ''}`}>
```
VeraZab: yeah, I think we're not really using that bem stuff, we should probably remove it at some point..

nicolaskruchten: This was added by @aulneau so we should get his input... we should either do one big pass where we remove it or add it everywhere in a single PR but agreed that we should standardize!

VeraZab: no : ) this was added before aulneau, we've just all been adding to it, but yeah @aulneau what do you think about that bem utility, useful?

VeraZab:
This is clean! I'm just going to check out this branch and test it out, but 💃

src/components/containers/Panel.js
nicolaskruchten: Oh, did you mean the `bem()` function or the actual BEM pattern?

And yes, the particular edit you're commenting on came from Thomas ;)

nicolaskruchten: actually come to think of it I was going to refactor it a bit, good that you chose this one to comment on!

nicolaskruchten: in the same family of functions as things like `bem()` I like https://github.com/JedWatson/classnames FWIW...

VeraZab: I meant the `bem` function, which was introduced before @aulneau started contributing to this project : )

PR #314, opened 2018-02-08
Multi-value support in trace tabs

nicolaskruchten:
This PR includes the change to hide text editors in multi-value contexts, and also adds support for styling text scatters, which we missed in the original passes through there :)


src/components/containers/derived.js
```diff
@@ -16,11 +16,7 @@ const AxesFold = connectAxesToLayout(Fold);
 
 const TraceTypeSection = (props, context) => {
   const {fullContainer} = context;
-  if (
-    fullContainer &&
-    fullContainer._fullInput &&
```
nicolaskruchten: I don't know why we were checking `_fullInput` here but it was failing in multi-valued contexts

src/components/fields/SymbolSelector.js
```diff
@@ -353,27 +354,38 @@ const SYMBOLS = [
...
 
-  setLocals(props) {
+  setLocals(props, context) {
     const {fullContainer} = props;
+    const {defaultContainer} = context;
 
     this.markerColor = nestedProperty(fullContainer, 'marker.color').get();
     this.borderWidth = nestedProperty(fullContainer, 'marker.line.width').get();
 
+    if (this.markerColor === MULTI_VALUED) {
```
nicolaskruchten: this is very tricky: we don't actually end up using default values here and I couldn't figure out how to do it... we use the values from the first trace. I've added a note to fit'n'finish

src/lib/multiValues.js
```diff
@@ -0,0 +1,77 @@
+import {MULTI_VALUED} from './constants';
```
nicolaskruchten: this was moved from `connectAxesToLayout` basically unmodified.

src/components/containers/derived.js
VeraZab: yeah _fullInput is where the true trace type for financial traces lie
_fullData.type for an ohlc trace will show scatter, _fullData._fullInput.type, will show ohlc

And all traces are supposed to have a _fullInput key, with type..so that's why I'm looking here
We could also look in Data.type..

But yeah, if you don't look there, it won't properly look for financials

src/components/fields/SymbolSelector.js
VeraZab: well, that's what the current workspace does, sounds ok to me!

src/components/containers/derived.js
nicolaskruchten: ah, right, I keep forgetting to check stuff with financial traces. right now they cause crashes apparently... joy!

VeraZab:
this is great @nicolaskruchten ! :dancer: once we put back _fullInput (in that comment above)

nicolaskruchten:
well, we can't just "put it back" as it doesn't work in all cases. need to find a more general solution. 

Seems like in general the financial traces cause a lot of problems :(

VeraZab:
what cases does it not work for?
I've just tried from master to create 2 ohlcs, that doesn't crash, there's some errors (trace.name and line.color), but it's mostly ok.

for fin charts, the true values are in _fullInput, that has to go back.
it's not just trace.type actually, it's also the true attributes, like {open, high, low, close}, you won't find those in gd._fullData

nicolaskruchten:
Per my comment, it doesn't work properly in grouped-trace contexts. I will fix it, don't worry, my point is I can't just undo my change :)

nicolaskruchten:
Ah, and the crash is just another variant of issues that crop up when we delete traces. It actually happens without financial traces, I just happened to have all financial traces up when I noticed it.

VeraZab:
well, Etienne admits that they're a bit of a mess, but that's what we've got now :(
https://github.com/plotly/plotly.js/issues/2281

src/default_panels/StyleTracesPanel.js
```diff
@@ -29,7 +29,7 @@ import {localize} from '../lib';
 
 const StyleTracesPanel = ({localize: _}) => (
   <TraceAccordion canGroup>
-    <TextEditor label={_('Name')} richTextOnly />
+    <TextEditor label={_('Name')} attr="name" richTextOnly />
     <TraceOrientation
```
VeraZab: oh yeah 👍

nicolaskruchten:
ok ready for re-review. I patched things up so that nothing is totally crazy but I've excluded the financial traces from the grouping altogether as it wasn't possible for me to find a constellation of settings that made them work correctly.

nicolaskruchten:
(I logged this in #290 of course for another pass later)

src/components/containers/derived.js
```diff
@@ -18,8 +18,9 @@ const TraceTypeSection = (props, context) => {
   const {fullContainer} = context;
   if (
     fullContainer &&
-    fullContainer._fullInput &&
-    props.traceTypes.includes(fullContainer._fullInput.type)
+    ((fullContainer._fullInput &&
+      props.traceTypes.includes(fullContainer._fullInput.type)) ||
```
nicolaskruchten: @VeraZab `_fullInput` is not available in some code paths, so this is the compromise. Specifically, in the grouped context, the all keys starting with `_` are missing from `container` as per the original logic from the grouped Axes. If I remove this exclusion, we end up with circular JSON objects that cannot be serialized :)

src/components/containers/TraceAccordion.js
```diff
@@ -52,31 +52,30 @@ class TraceAccordion extends Component {
...
+      const traceType = plotlyTraceToCustomTrace(nextTrace);
+      if (!allTraces[traceType]) {
+        allTraces[traceType] = [];
+      }
+      allTraces[traceType].push(index);
+      return allTraces;
+    }, {});
+
+    const groupedTraces = Object.keys(tracesByGroup)
+      .filter(traceType => !['ohlc', 'candlestick'].includes(traceType))
```
nicolaskruchten: this is ugly but prevents really weird outcomes.

src/components/containers/TraceAccordion.js
```diff
@@ -52,31 +52,30 @@ class TraceAccordion extends Component {
...
+      .map((traceType, index) => {
```
VeraZab: ok!

VeraZab:
great :dancer:

PR #319, opened 2018-02-09
Added x | y | z | error bars

VeraZab:
@nicolaskruchten ready for a look
ex


nicolaskruchten:
UI looks good, but I'm puzzled about the relationship to the Workspace UI... Workspace seems to have two dropdowns per error bar?

image

VeraZab:
yeah, I didn't quite get what the second one was doing, but now I get it, it's the 'arrayminus' key.. not sure what to name that, but ok, adding it in.

VeraZab:
ok should be good now @nicolaskruchten

nicolaskruchten:
Hmmm OK so digging into the plotly.js docs there are actually a bunch more options here. If you don't specify an array then the "type" is "percent" it looks like, but we should be able to specify the percentage. Should we add the controls for "type" and "value" and "symmetric" in here? It's weird to just have the two arrays... (I know the workspace doesn't do this, but the editor framework makes it pretty easy ;)

nicolaskruchten:
`valueminus` and `arrayminus` could appear only if `symmetric` is false, and the `type` could control if we see `value*` or `array*`

VeraZab:
Yeah, there's lots of options that why I felt like it might be nice to have a fold in axes for these..
But then having DataSelectors in folds is a little against the current organization of things..

Ok well, I'll see how it looks if I just add them all here, maybe it's not too much after all

nicolaskruchten:
I hear ya, but I think they won't all appear at once... worst case is you have:

* show: true
* type: value
* symmetric: false
* value: 
* valueminus

And if we want we can probably merge show and symmetric into one custom control with three options: hide, symmetric, asymmetric

VeraZab:
ok I'll see how that looks, maybe i'll even have 3 separate sections one per axis

VeraZab:
yeah, I don't know what's up.. I have to click twice on those RadioBlock options for them to be selected, although they're controlled by component state.. 

have to look more into that

nicolaskruchten:
Is this ready for review?

VeraZab:
no not yet

nicolaskruchten:
OK cool. I like keeping "[WIP]" in the PR description for this kind of thing, so that everyone knows it's not ready yet :)

VeraZab:
updated controls:

screen shot 2018-02-12 at 3 03 11 pm

screen shot 2018-02-12 at 10 50 24 am

screen shot 2018-02-12 at 10 54 27 am

screen shot 2018-02-12 at 3 02 53 pm

screen shot 2018-02-12 at 3 03 18 pm

screen shot 2018-02-12 at 3 03 01 pm

screen shot 2018-02-12 at 3 03 26 pm

VeraZab:
cc @nicolaskruchten now ready for review : )

nicolaskruchten:
💃

PR #323, opened 2018-02-13
fixing tests

nicolaskruchten:
This seems needed to make tests pass after an `rm -rf node_modules && yarn install`... /cc @aulneau somehow the entry in `.babelrc` was messing up the tests with up-to-date dependencies.


VeraZab:
💃

PR #324, opened 2018-02-13
Temporary fix for AxesSelector

nicolaskruchten:
This fixes the issue where the Axes panel is empty for 3d traces. Needs another pass to set it to the first non-all setting by default (already added to #290), but this lets us move forward.


VeraZab:
💃  thanks!

PR #331, opened 2018-02-14
Axes for all

VeraZab:
- [ ] xref/yref in annotations (next pr)


src/default_panels/StyleAxesPanel.js
```diff
@@ -19,313 +20,317 @@ import {localize} from '../lib';
...
 const StyleAxesPanel = ({localize: _}) => (
   <TraceRequiredPanel>
-    <AxesFold name={_('Titles')}>
-      <TextEditor attr="title" />
-      <FontSelector label={_('Typeface')} attr="titlefont.family" />
-      <Numeric label={_('Font Size')} attr="titlefont.size" units="px" />
-      <ColorPicker label={_('Font Color')} attr="titlefont.color" />
-    </AxesFold>
+    <AxisRequiredPanel
+      emptyPanelHeader={_('Your plot does not have any axes to style.')}
```
VeraZab: I just wrapped everything in AxisRequiredPanel here

src/lib/connectAxesToLayout.js
```diff
@@ -35,18 +33,41 @@ export default function connectAxesToLayout(WrappedComponent) {
...
-      if (plotly) {
-        this.axes = plotly.Axes.list(graphDiv);
-      } else {
-        this.axes = [];
-      }
+
+      this.axes = [];
+
+      // Plotly.js should really have a helper function for this, but until it does..
+      Object.keys(fullContainer._subplots)
```
VeraZab: ok, so subplots will be step 2, after this pr, that is why I disabled a test somewhat related to them, above.
Need to think about this a little more when I'm writing this function here.

VeraZab:
@nicolaskruchten this is a first pass that takes care of axes for all plot types, and does not use the plotly.Axes.list() function that doesn't work on plots other than cartesian.

I'm taking care of subplots next..

Would you give this a look?

nicolaskruchten:
💃

PR #334, opened 2018-02-15
[WIP] Subplots

VeraZab:
- [ ] SubplotTracePicker has to only show compatible traces as options
- [ ] Calculate better subplot defaults when adding a subplot
- [ ] Subplot folds dissappear when clicking away from the panel, maybe state needs to be higher, maybe I need to write them to gd.
- [ ] Check multi trace assignment
- [ ] boundary controls have to be attached to their actual containers if I don't want to create an EDITOR_ACTION for control updates

### DIFF
--- a/src/DefaultEditor.js
+++ b/src/DefaultEditor.js
@@ -4,6 +4,7 @@ import {PanelMenuWrapper} from './components';
 import {localize} from './lib';
 import {
   GraphCreatePanel,
+  GraphSubplotsPanel,
   StyleLayoutPanel,
   StyleAxesPanel,
   StyleLegendPanel,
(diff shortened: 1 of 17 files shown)


VeraZab:
@alexcjohnson or @etpinard or @monfera, as I'm playing around with subplots I notice that for some chart types `.domain` is in gd.data, vs others in gd.layout:

domain:
**in data**
pie.domain
parcoords.domain
sankey.domain
table.domain

**in layout**
layout.xaxis.domain
layout.yaxis.domain
layout.ternary.domain
layout.scene.domain
layout.geo.domain
layout.mapbox.domain
layout.polar.domain
layout.radialaxis.domain
layout.angularaxis.domain

Is it because the trace types that have it in gd.data are considered to have no axes? Or was there some other reason. Trying to find generalizations here. Thanks!

monfera:
@VeraZab yes, something similar, I think it boils down to Cartesian-like vs. not. Technically, `parcoords` does have axes, but it's a bunch of parallel axes and there's currently no overlay, ie. one can't superimpose multiple parcoords atop of one another. It is on the other hand possible to have multiple elements in the `data` array that are plotted in different places, controlled via the `domain` in `data[i]`. So the difference, I believe, is that axes - if present at all - are controlled at the `data` level in the first group and jointly, at the `layout` level in the second group.

alexcjohnson:
I guess  you could say these trace types have no axes - I think of it as whether the trace lives inside a subplot, in which case the subplot needs to describe its placement within the plot with a `domain`, vs the trace standing on its own, in which case it's the *trace* that needs the `domain`.

Every subplot (and cartesian axis) needs a `domain`.
Every trace either has a subplot attribute or attributes (`subplot`, `scene`, `geo`, or `xaxis` and `yaxis`) or it has its own `domain`.

PR #339, opened 2018-02-20
bumping react-plotly.js version

nicolaskruchten:
Should resolve some EventEmitter noise in the console :)


VeraZab:
nice thanks!

Issue #341, opened 2018-02-21
Strip out tags in title displays


nicolaskruchten:
We display trace names as Fold titles and axis names in dropdowns, and we should strip out tags when we do, perhaps using https://www.npmjs.com/package/striptags

![image](https://user-images.githubusercontent.com/203523/36485364-0b5d9f30-16ea-11e8-9cd8-42de416fb1ce.png)

![image](https://user-images.githubusercontent.com/203523/36485388-1be31d30-16ea-11e8-80bb-910ff7909bb4.png)

nicolaskruchten:
@VeraZab what do you think of https://www.npmjs.com/package/striptags ? Do we have a different library we use for this sort of thing? Something built in to plotly.js?

VeraZab:
Striptags looks good, let's use that if plotly.js doesn't have anything built in. (cc @alexcjohnson we don't have anything in plotly.js doing this correct?)

alexcjohnson:
We do have [`svgTextUtils.plainText`](https://github.com/plotly/plotly.js/blob/master/src/lib/svg_text_utils.js#L273) - which has the distinction of only stripping the tags that plotly.js actually supports. I *think* that means the results will be more consistent with what you see on the actual plot, in case someone uses an unsupported tag.

VeraZab:
oh nice! Then I think it'd be a good idea @nicolaskruchten to use svgTextUtils. All the places we allow the user to use html tags is because it's an input to plotly.js afaik..

nicolaskruchten:
Hmmm, this seems less safe than the NPM module which isn't regex-based...

alexcjohnson:
> Hmmm, this seems less safe than the NPM module which isn't regex-based...

Does that mean there's a bug in plotly.js? That we should fix (possibly by using the npm module) and then use it here too?

nicolaskruchten:
I don't know if there's a bug in plotly.js, just that regex-based tag sanitizers are sometimes considered less safe than simpler state-machine-based ones... See this for example: https://stackoverflow.com/questions/24723/best-regex-to-catch-xss-cross-site-scripting-attack-in-java/535022#535022

nicolaskruchten:
For example:

`plainText("<<a>script>alert('ciao');<</a>/script>")` 
returns 
`'< script>alert(\'ciao\');< /script>'`

alexcjohnson:
[...]

Issue #340, opened 2018-02-20
XSS vector in trace names


nicolaskruchten:
Here: https://github.com/plotly/react-plotly.js-editor/blob/15c29f58065f57837fc3f08c1d1c3ea436d505d1/src/components/containers/Fold.js#L101

PR #342, opened 2018-02-21
fix 2 XSS vulns

nicolaskruchten:
This fixes a vulnerability in our editor and removes the use of `dangerouslySetInnerHTML` per #340 ... Created #341 to make things look nicer after this is merged.


VeraZab:
:dancer:

PR #345, opened 2018-02-22
strip out tags in display

### LINKED ISSUE #341: Strip out tags in title displays

nicolaskruchten:
Fixes #341


VeraZab:
that's ready for review right?
looks like : ) 💃

Issue #280, opened 2018-01-25
LayoutNumericFraction and Inverse and lighting controls should be sliders


nicolaskruchten:
For workspace parity.

VeraZab:
and box width #92

VeraZab:
just a small component to make and add, after the chart picker Modal : )

PR #346, opened 2018-02-22
numeric slider

nicolaskruchten:


nicolaskruchten:
@aulneau do you think you could help me out with the CSS here please? Here's a side-by-side of old vs new... we don't have to get all the way to 'old' but partway would be nice :) I already copied the default styles and hacked them somewhat to get from huge/green to small/ grey ;)

image

src/components/fields/derived.js
```diff
@@ -105,14 +105,45 @@ export const AxesRange = connectToContainer(UnconnectedNumeric, {
...
+    }
+  };
+  plotProps.max = 100;
+  plotProps.min = 0;
+};
+
+export const NumericFraction = connectToContainer(UnconnectedNumericFraction, {
+  modifyPlotProps: numericFractionModifyPlotProps,
+});
+
```
VeraZab: ok, so this is my general control for when I want a slider instead of a standard numeric.

VeraZab:
looks good 💃 after stylings!

src/components/fields/derived.js
nicolaskruchten: only if you want it to be expressed as a percentage! if your min is 2 and your max is 4 then 50% = 3 etc... if you want a slider from 2 to 3 you should use `<Numeric showSlider ... >`

nicolaskruchten:
@aulneau (bump) ... do you know if you'd have any time this week to help with the CSS on this one? Once done this resolves #280 btw :)

aulneau:
@nicolaskruchten I can check this out tonight / tomorrow AM!

src/styles/components/fields/_field.scss
```diff
@@ -40,6 +40,7 @@
   }
   &__title {
     width: 30%;
+    //flex-shrink:0;
```
aulneau: This should be enabled. The titles should all be the same width, but if I keep this on, certain tabs will escape the boundaries of the content area of the field. @nicolaskruchten @VeraZab what do you two think of changing these fields that have the radio buttons (3+) into a label on top, and then a full width field widget?

flex-shrink: 1
screen shot 2018-02-27 at 1 23 31 pm

flex-shrink: 0
screen shot 2018-02-27 at 1 23 41 pm

nicolaskruchten: I'm OK with your proposal.

nicolaskruchten:
These look good except they're too tall... Ideally a numeric with a slider should be the same height as one with the arrows, because in some cases the UI can switch from one to the other and we don't want a vertical jump in that case...

image

aulneau:
Whoops, sorry. there was some default margins in there :)

should be good to go!

screen shot 2018-02-27 at 1 58 01 pm

PR #357, opened 2018-02-23
Subplots2

VeraZab:


VeraZab:

src/components/fields/AxisCreator.js
```diff
@@ -0,0 +1,155 @@
+import Dropdown from './Dropdown';
+import Field from './Field';
+import Info from './Info';
+import PropTypes from 'prop-types';
+import React, {Component, Fragment} from 'react';
+import {connectToContainer, localize, traceTypeToAxisType} from 'lib';
+import {EDITOR_ACTIONS} from 'lib/constants';
+import {PlusIcon} from 'plotly-icons';
+
```
nicolaskruchten: this component should show the axis title, not just the `x`, `x2`, `y2` axis IDs

nicolaskruchten:
This looks pretty good! A couple of my comments need addressing... I've also noticed some odd behaviour in the UI itself but nothing specific I can point to as a bug just yet. Mostly it seems hard or impossible to build a basic dual-Y-axis chart with this UI unfortunately. Maybe I'm missing something I'll try to do some more systematic review after lunch. Sorry to be out of sync today.

VeraZab:
hmm, i did set the y2 axis to be easy to make..
ex

and i made axes name reflect the names they're given by the user:
ex

src/components/fields/derived.js
```diff
@@ -4,12 +4,34 @@ import {UnconnectedFlaglist} from './Flaglist';
...
 import {
+  capitalize,
   connectLayoutToPlot,
   connectToContainer,
+  getAllAxes,
   supplyLayoutPlotProps,
   striptags,
 } from 'lib';
 
+export const AxisAnchorDropdown = connectToContainer(UnconnectedDropdown, {
```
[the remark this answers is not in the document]
VeraZab: https://github.com/plotly/react-plotly.js-editor/issues/360

src/components/fields/Field.js
```diff
@@ -58,6 +61,14 @@ class Field extends Component {
             </MenuPanel>
           ) : null}
         </div>
+        {action ? (
```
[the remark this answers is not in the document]
VeraZab: well, it may :) if you don't mind too much I'd like to keep it in here for the moment

src/components/fields/AxisCreator.js
VeraZab: that's done, as per below gif

src/lib/__tests__/multiValued-test.js
```diff
@@ -7,7 +7,7 @@ import {connectLayoutToPlot, connectAxesToLayout} from '..';
 import {mount} from 'enzyme';
 
 describe('multiValued Numeric', () => {
-  it.only('uses placeholder and empty string value', () => {
```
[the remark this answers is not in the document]
VeraZab: this is just a test I uncommented, because my previous merge was messing up this behaviour and I knew I'd fix it by introducing subplots..

src/components/fields/AxisCreator.js
nicolaskruchten: It's done in the `AxesSelector` but not the `AxisCreator` right?

nicolaskruchten:
So if I create two traces and then set one of them to the second Y-axis, I only ever see 1 trace displayed, I'm not sure where the other one goes!

src/lib/__tests__/multiValued-test.js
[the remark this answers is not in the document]
nicolaskruchten: yeah, I meant that this kind of `it.only` should never be allowed right? like we don't allow `fit` and `xit` etc?

VeraZab:
> I only ever see 1 trace displayed, I'm not sure where the other one goes!
are you talking about the example in my gif, or some that you've tried?

src/components/fields/Field.js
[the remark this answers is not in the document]
VeraZab: or looking at it again, yeah, I may change this, it's a bit ugly

src/components/fields/AxisCreator.js
VeraZab: right..

nicolaskruchten:
I'm trying it locally. If I set up trace 0 with x=col1, y=col1, then trace1 with x=col3, y=col3, then I set trace1's y-axis to a new axis, I see the following... where are the blue points for trace0?

image

src/components/fields/Field.js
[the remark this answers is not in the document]
nicolaskruchten: well, it's likely not worth going through the effort of moving it now, but in future I'd be more comfortable adding in this kind of thing higher up in the stack, not in the base-level components like `Field` :)

nicolaskruchten:
Ah ok, it looks like `yaxis2.overlaying = 'y'` is needed... Not sure if forcing that on creation makes sense or not. We're already moving it to the right side right?

VeraZab:
@alexcjohnson would you take a look at this pr for me? see if the logic makes sense. 

Also, about creating a second y axis, I haven't noticed that part of my data disappears when I assign yaxis: 'y2' to a trace..
https://codepen.io/plotly/pen/648434b7b06688fc6b00067592d616b0
As Nicolas points out above, it seems like an `overlaying: 'y'` is needed for both traces to show up. Is this something that maybe plotly.js should be handling by default? 

If it's not something to be handled by default by plotly.js, what's the best update that I need to send to plotly.js for all my traces to continue to show up after I've created new axes?

src/lib/__tests__/multiValued-test.js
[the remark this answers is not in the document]
VeraZab: yeah it should not be allowed..

alexcjohnson:
>  it seems like an `overlaying: 'y'` is needed for both traces to show up. Is this something that maybe plotly.js should be handling by default?

That would be tricky for plotly.js to figure out - we'd have to infer that you intend the two y axes to have not only the same domain but to actually be used within the same subplot (as opposed to just side-by-side plots). Theoretically possible, but I feel like it would be hard to make robust. Whereas here when you're making the second axis, you know whether the intent is overlaid, side-by-side, stacked, or whatever, and you can send `yaxis2.overlaying='y'` if that's the intent.

nicolaskruchten:
@alexcjohnson I'm having trouble understanding the `overlay: free` behaviour... if y and y2 are both anchored to x, and both are overlay: free (the default) then why don't both traces show up on the plot?

src/components/containers/ImageAccordion.js
```diff
@@ -13,8 +13,8 @@ class ImageAccordion extends Component {
 
     const content =
       images.length &&
-      images.map((ann, i) => (
-        <ImageFold key={i} imageIndex={i} name={ann.text} canDelete={canAdd}>
+      images.map((img, i) => (
+        <ImageFold key={i} imageIndex={i} name={img.text} canDelete={canAdd}>
```
alexcjohnson: images don't have `.text` do they? Perhaps make a name from `.source`? Not sure what to do about shapes... index and type perhaps?

nicolaskruchten: hehe so this change is @VeraZab cleaning up my copy-paste work from a previous PR... When I created the Images panel I'd based it off of the Notes panel but not renamed `ann` to `img` (and ditto Shapes). So any issues with this code should probably go in a separate issue that I can handle ;)

alexcjohnson:
>  I'm having trouble understanding the `overlay: free` behaviour...

Heh that's because there isn't any `free` behavior. That's a mistake in the attribute definition, I'll fix it. If you look at `_fullLayout` the only options you should actually see are an axis id, `undefined` (there are no possible axes to overlay) or `false` (you declined to overlay another axis).

`overlaying` does two related but things when you enable it: 1) it forces the axis to copy the domain of the axis it's overlaying, and 2) it draws its data (when sharing an x axis too) on top of the data for the axis it's overlaying, without first covering the lower subplot in an opaque background.

It's helpful to remember the case you DO want one subplot to block another: inset plots. You do not want data from the main plot to show up behind the inset. I'm sure there could be smarter ways to handle this kind of case, since it's admittedly rare... but hopefully that at least makes it clear how we arrived at the situation we have now.

nicolaskruchten:
Ahh, thanks for the explainer, it makes sense now :)

src/lib/getAllAxes.js
```diff
@@ -0,0 +1,39 @@
...
+      .forEach(type => {
+        fullLayout._subplots[type].forEach(subplot => {
+          if (['xaxis', 'yaxis'].includes(type)) {
+            // subplot will look like x2, x45, convert it to xaxis2, xaxis45
+            subplot = // eslint-disable-line no-param-reassign
+              subplot.length > 1
+                ? subplot.slice(0, 1) + 'axis' + subplot.slice(1)
+                : subplot + 'axis';
+
+            fullLayout[subplot].subplot = subplot;
```
alexcjohnson: If you're adding something to `fullLayout` that's not a real attribute, can you make it private, ie `_subplot`?

src/lib/connectAxesToLayout.js
```diff
@@ -2,14 +2,19 @@ import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import nestedProperty from 'plotly.js/src/lib/nested_property';
 import {deepCopyPublic, setMultiValuedContainer} from './multiValues';
-import {getDisplayName, localize, capitalize} from '../lib';
+import {capitalize, getAllAxes, getDisplayName, localize} from '../lib';
```
alexcjohnson: Also not new in this PR, but Isn't this `from '.'`? Actually, wouldn't it be better to import from the individual files anyway rather than `lib/index`, since this way is a circular import? It obviously doesn't break, but makes me uneasy...

src/lib/constants.js
```diff
@@ -48,3 +48,19 @@ export const RETURN_KEY = 'Enter';
...
+    'heatmap',
+    'contour',
+    'ohlc',
+    'candlestick',
+  ],
+  ternary: ['scatterternary'],
+  gl3d: ['scatter3d', 'surface', 'mesh3d'],
+  geo: ['scattergeo', 'choropleth'],
+  mapbox: ['scattermapbox'],
+};
```
alexcjohnson: This feels fragile to me - an alternative that wouldn't require any hardcoding here would be to look at `fullTrace._module.basePlotModule.name`. Traces that don't have a subplot, like `pie`, will define their own `basePlotModule`, so then you can filter on names in the list of known subplot types.

src/components/fields/derived.js
```diff
@@ -38,6 +38,29 @@ export const AxisAnchorDropdown = connectToContainer(UnconnectedDropdown, {
   },
 });
 
+export const AxisOverlayDropdown = connectToContainer(UnconnectedDropdown, {
+  modifyPlotProps: (props, context, plotProps) => {
+    let options = [];
+    if (plotProps.fullContainer.subplot.includes('xaxis')) {
```
nicolaskruchten: could this be DRYed up with a map on `xaxis, yaxis` ?

src/lib/__tests__/multiValued-test.js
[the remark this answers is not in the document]
nicolaskruchten: #362 will prevent you from trying this trick again ;)

src/lib/constants.js
VeraZab: I'll add this to our list of refactorings in fit n' finish.. thanks..

VeraZab:
@nicolaskruchten would you checkout this branch once more and see if it looks ok to you..?

nicolaskruchten:
I'm really sorry that this one keeps dragging on and on but I'm seeing errors in the Axes panel after creating a second Y axis, and some weird behaviour after creating a second X axis ... :(

nicolaskruchten:
OK there's still a subtle bug which seems to only appear when you create new axes, then click on Axes panel, then come back and create even more new axes. Sometimes I get this output:

image

nicolaskruchten:
Here's a way to get the same stack trace:

1. create trace0: x=col1, y=col1
2. create trace1: x=col2, y=col2
3. create new y axis for trace1
6. create new x axis for trace0
7. `factory.js?b9b2:128 Error while plotting: TypeError: Cannot read property 'overlays' of undefined`

alexcjohnson:
> OK there's still a subtle bug which seems to only appear when you create new axes, then click on Axes panel, then come back and create even more new axes. Sometimes I get this output:

If you can replicate this as `restyle`/`relayout`/`react` calls to plotly.js I'd love to take a look. It's conceivable that https://github.com/plotly/plotly.js/pull/2416 fixes it but also quite possible it's unrelated.

nicolaskruchten:
@alexcjohnson I'll see what I can do, this one came out of some button-mashing approach where I sort of clicked back and forth across various views and hammered on the +Axis buttons over and over without keeping a mental stack of how to replicate ;)

nicolaskruchten:
Note that that orange trace is really composed of three co-linear points, so I'm really puzzled to see it rendered this way!

alexcjohnson:
It looks to me as though you're seeing two copies of the orange trace, one that should have been cleaned up when its subplot was rendered obsolete and one that actually belongs.

nicolaskruchten:
Ah interesting. This might be a race condition between plotly.js and React then as well.

nicolaskruchten:
To capture the convo we just had with @VeraZab ... the following UX change makes sense to me: for a given trace, the +Axis button should be disabled for a given axis if that trace is the only one currently using that axis. I.e. in a 2-trace case, if each trace is on its own x-axis, the user shouldn't be able to create a new x-axis. In a 3-trace case where traces A and B are on x1 and trace C is on x2, then the +Axis button should be disabled for the x-axis for trace C, as if it was used, it would orphan x2. This means that this check is not the same as "number of axes = number of traces"

src/lib/getAllAxes.js
```diff
@@ -0,0 +1,75 @@
...
+    Object.keys(fullLayout._subplots)
+      .filter(
+        // xaxis and yaxis already included separately in _fullLayout._subplots
+        type => type !== 'cartesian' && fullLayout._subplots[type].length !== 0
+      )
+      .forEach(type => {
+        fullLayout._subplots[type].forEach(subplot => {
+          if (['xaxis', 'yaxis'].includes(type)) {
+            // subplot will look like x2, x45, convert it to xaxis2, xaxis45
+            subplot = // eslint-disable-line no-param-reassign
```
nicolaskruchten: why disable eslint here as opposed to just creating a new variable?

nicolaskruchten:
💃 💃 ! thanks for putting up with my million comments on this one :)

Issue #358, opened 2018-02-23
axes panel scroll


VeraZab:
currently on master
https://plotly.github.io/react-plotly.js-editor/
![screen shot 2018-02-23 at 11 47 37 am](https://user-images.githubusercontent.com/4257572/36605687-6cade230-188f-11e8-97e1-04c0b193ccbd.png)

@aulneau would you please take a look..?

PR #368, opened 2018-03-03
Scrollbar

VeraZab:
fixes: #358
screen shot 2018-03-02 at 11 33 41 pm

This was displaying weirdly because TraceRequiredPanel was returning a Panel and so was AxisRequiredPanel. So there were two panels wrapped in one another.

I refactored this code a little so that AxisRequiredPanel is not needed and instead we can add extra conditions to the TraceRequiredPanel, along with extraEmptyPanel messages.


nicolaskruchten:
Remind me again under which conditions it's possible to have traces defined but not axes? Just pie charts?

VeraZab:
pie, mapbox

nicolaskruchten:
💃

PR #370, opened 2018-03-05
Add table trace

VeraZab:
@nicolaskruchten do you want to take a look on this one?

I think there's some plotly.js issues to log.. for hoverinfo, and fill.color (when you set a color and erase, it goes to black instead of defaulting to white..), and there's the table icon to add to plotly-icons.. but all this can be done outside of here

closes: #335

### DIFF
--- a/dev/App.js
+++ b/dev/App.js
@@ -10,9 +10,12 @@ import Nav from './Nav';
 import ACCESS_TOKENS from '../accessTokens';
 
 const dataSources = {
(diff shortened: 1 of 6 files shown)


nicolaskruchten:
the little trace icon is still 'line'... other than that 💃 

image

VeraZab:
yeah, it's line because we don't have that icon in plotly-icons yet..
we have to add it to plotly-icons, will do!

VeraZab:
merging this branch though for now, as this is not something to be fixed here..

PR #371, opened 2018-03-05
Remove dep to fix build scripts

aulneau:
I was importing some css from `node_modules` and the scripts we use to generate css can't read from `node_modules` the same way scss can. I removed the dependency and just added it manually as it was very small (albeit we aren't currently using the tooltips, so it's still subject to change).

This fixes the build errors that was causing.


VeraZab:
thanks 💃

PR #373, opened 2018-03-05
Move svg's to aws for now

aulneau:
This PR is a temp fix to move our new SVG icons for the advanced trace selector to aws until we can convert them to react components.


VeraZab:
those svgs take time to load.. looks a little ugly.. 
closing..
we'll use old icons until #375 is in

PR #377, opened 2018-03-06
better fix for deletion bugs

nicolaskruchten:


src/components/containers/Fold.js
```diff
@@ -84,26 +82,29 @@ class Fold extends Component {
...
-      doDelete ? (
+    const icon = Icon ? <Icon className="fold__top__icon" /> : null;
+
+    const deleteButton =
+      canDelete && typeof deleteContainer === 'function' ? (
         <div
           className="fold__top__delete js-fold__delete"
-          onClick={handleClick}
+          onClick={e => {
+            e.stopPropagation();
```
nicolaskruchten: this is the important line: when we click on the delete button, we do NOT want the Fold-folding code to fire, otherwise React may rerender with the graphDiv in an inconsistent state

VeraZab:
💃

PR #380, opened 2018-03-06
Rename react-plotly.js-editor to react-chart-editor

VeraZab:


VeraZab:
@bpostlethwaite there's probably something on the circle side of things that I need to do as well right? if I've renamed the gh repo..?

PR #383, opened 2018-03-06
Refactor

VeraZab:
@nicolaskruchten what do you think of this refactor?


nicolaskruchten:
it's a step in the right direction! 💃

Issue #321, opened 2018-02-12
full loop tests


nicolaskruchten:
Our current testing framework only really tests things "one way" in that we fixture up a figure, get plotly.js to process it, render the editor, and optionally check that the right event payloads fire on input, but we don't actually send those payloads back into plotly.js to close the loop. This prevents us from doing real end to end tests and allows bugs like the deletion one I hunted all of last week to evade automated tests.

Let's add some test utils to make end to end testing easier and then actually start building such tests!

PR #384, opened 2018-03-07
Change default export to combined EditorControls + Plot component

nicolaskruchten:
I'm happy to rename the new component but the idea here is to provide something that, though it keeps a gross DOM node in state, looks and behaves like a dumb component which bundles an editor and a plot area.


dev/App.js
```diff
@@ -22,18 +21,13 @@ const dataSourceOptions = Object.keys(dataSources).map(name => ({
...
   constructor() {
     super();
 
-    // The graphDiv object is passed to Plotly.js, which then causes it to be
-    // overwritten with a full DOM node that contains data, layout, _fullData,
-    // _fullLayout etc in handlePlotUpdate()
     this.state = {
-      graphDiv: {},
-      plotRevision: 0,
+      data: [],
```
nicolaskruchten: note the new state in `App`: just data and layout!

dev/App.js
```diff
@@ -49,61 +43,36 @@ class App extends Component {
...
-          </div>
-        </div>
+        <PlotlyEditorWithPlot
+          data={this.state.data}
+          layout={this.state.layout}
+          config={{mapboxAccessToken: ACCESS_TOKENS.MAPBOX, editable: true}}
+          dataSources={dataSources}
+          dataSourceOptions={dataSourceOptions}
+          plotly={plotly}
+          onUpdate={(data, layout) => this.setState({data, layout})}
```
nicolaskruchten: note the new update call: just data and layout!

src/PlotlyEditor.js
```diff
@@ -69,7 +69,7 @@ class PlotlyEditor extends Component {
           this.props.afterUpdateTraces(payload);
         }
         if (this.props.onUpdate) {
-          this.props.onUpdate();
+          this.props.onUpdate(graphDiv.data, graphDiv.layout);
```
nicolaskruchten: in this iteration, we mutate and pass along. In the next iteration, we'll use `immutability-helper` to NOT mutate and pass along copies :)

src/PlotlyEditorWithPlot.js
```diff
@@ -0,0 +1,58 @@
+import React, {Component} from 'react';
+import createPlotComponent from 'react-plotly.js/factory';
+import PlotlyEditor from '../src';
+import PropTypes from 'prop-types';
+
+class PlotlyEditorWithPlot extends Component {
+  constructor(props) {
+    super();
+    this.state = {graphDiv: {}};
```
nicolaskruchten: The new component only stores the graphDiv in state, which it gets from Plot.

So the lifecycle of a user edit is now:
* `data`/`layout` from `PlotlyEditor` up to `PlotlyEditorWithPlot` up to `App`
* `setState` causes `render` for `App`
*  `data`/`layout` from `App` down to `PlotlyEditorWithPlot` down to `Plot`
* `graphDiv` from `Plot` up to `PlotlyEditorWithPlot`
* `setState` causes `render` for `PlotlyEditorWithPlot`
* `graphDiv` from `PlotlyEditorWithPlot` down to `PlotlyEditor`

bpostlethwaite:
here is the batch reporter end-to-end test https://github.com/plotly/plotly.js-batch-reporter/blob/master/example/src/index.test.js 

There are a few accompanying helper functions. One of which https://github.com/plotly/plotly.js-batch-reporter/blob/master/example/src/index.test.js#L11 may be instructive.

bpostlethwaite:
also since things are async I use `await` https://github.com/plotly/plotly.js-batch-reporter/blob/master/example/src/index.test.js#L77 plus an async type of `waits_for` routine https://github.com/plotly/plotly.js-batch-reporter/blob/master/example/src/lib/testUtils.js#L78

VeraZab:
This is what we suspected could make the editor really slow right? And it isn't really slow?
For names: `Editor`, `Plot`, `PlotWithEditor`?
But this looks good to me!

nicolaskruchten:
@VeraZab Well, I didn't think it would make things "really slow" I said there might be a performance hit from extra renders. It doesn't feel bad to me at all though :)

nicolaskruchten:
Actually the bigger performance hit probably came from #381

nicolaskruchten:
@VeraZab @bpostlethwaite here's my more radical proposal: let's change the top-level export to be this new combined component that looks more React-idiomatic. I've renamed the top-level component to `PlotlyEditor` and the old one to `EditorControls` in this PR.

nicolaskruchten:
If we like it, I can fix all the other examples and write some tests against it. I'm pretty convinced that this is a better way forward but we can discuss in person when @VeraZab  is back :)

VeraZab:
I like it!

nicolaskruchten:
OK so remaining to do on this PR:

* [ ] add an end to end testing framework stub
* [ ] fix the README
* [ ] fix the existing examples

nicolaskruchten:
@VeraZab check out how nice the examples are now! `simple` and `redux` :)

examples/demo/src/App.js
```diff
@@ -42,61 +41,36 @@ class App extends Component {
...
+          layout={this.state.layout}
+          config={config}
+          dataSources={dataSources}
+          dataSourceOptions={dataSourceOptions}
+          plotly={plotly}
+          onUpdate={(data, layout) => this.setState({data, layout})}
+          useResizeHandler
+          debug
+          advancedTraceTypeSelector
+        />
```
VeraZab: yeahhh!! I like it!! 🎉

package.json
```diff
@@ -20,7 +20,7 @@
     "raf": "^3.4.0",
     "react-color": "^2.13.8",
     "react-colorscales": "^0.4.2",
-    "react-plotly.js": "^1.6.0",
+    "react-plotly.js": "^1.7.0",
     "react-rangeslider": "^2.2.0",
     "react-select": "^1.0.0-rc.10",
     "react-tabs": "^2.2.1",
```
VeraZab: on ajoute le nouveau plotly.js aussi?

nicolaskruchten: sure

src/PlotlyEditor.js
```diff
@@ -1,280 +1,67 @@
...
-          ' plotly-editor--theme-provider' +
-          `${this.props.className ? ` ${this.props.className}` : ''}`
-        }
-      >
-        <ModalProvider>
-          {this.props.graphDiv &&
-            this.props.graphDiv._fullLayout &&
-            (this.props.children ? this.props.children : <DefaultEditor />)}
-        </ModalProvider>
+      <div className="app">
```
vdh: Should `app` here be replaced with a BEM class instead? (Perhaps by adjusting/moving the one now in `EditorControls`?)

nicolaskruchten: Good point, I just copied stuff over but `app` is a terrible class to bake in ;)

nicolaskruchten:
@VeraZab @bpostlethwaite ready for review. The end-to-end test is a stub but it works and can be built upon in future PRs :)

partly addresses #321

src/PlotlyEditor.js
```diff
@@ -1,280 +1,66 @@
...
+          locale={this.props.locale}
+          traceTypesConfig={this.props.traceTypesConfig}
+          dictionaries={this.props.dictionaries}
+        >
+          {this.props.children}
+        </EditorControls>
+        <div
+          className="plotly_editor_plot"
+          style={{width: '100%', height: '100%'}}
+        >
```
VeraZab: could these encoded styles go in the plotly_editor_plot className?

src/PlotlyEditor.js
```diff
@@ -1,280 +1,66 @@
...
+          <this.PlotComponent
+            data={this.props.data}
+            layout={this.props.layout}
+            config={this.props.config}
+            useResizeHandler={this.props.useResizeHandler}
+            debug={this.props.debug}
+            onInitialized={graphDiv => this.setState({graphDiv})}
+            onUpdate={graphDiv => this.setState({graphDiv})}
+            style={{width: '100%', height: '100%'}}
+          />
```
VeraZab: and this too, can we not hard code here?

VeraZab:
:dancer:

bpostlethwaite:
:dancer:

Issue #379, opened 2018-03-06
PropType issue in TraceTypeSelector


nicolaskruchten:
Heads-up @VeraZab @aulneau 

```
Warning: Failed prop type: Invalid prop `actions` of type `function` supplied to `Item`, expected `array`.
    in Item (created by TraceTypeSelector)
    in TraceTypeSelector (created by LocalizedTraceTypeSelector)
    in LocalizedTraceTypeSelector (created by ModalProvider)
    in ModalProvider (created by PlotlyEditor)
    in div (created by PlotlyEditor)
    in PlotlyEditor (created by App)
    in div (created by App)
    in div (created by App)
    in App (created by HotExportedApp)
    in AppContainer (created by HotExportedApp)
    in HotExportedApp
```

aulneau:
I'll tack this fix on with my icons pr later today

VeraZab:
thanks @aulneau !

PR #387, opened 2018-03-07
Add in complex trace type svg components

aulneau:
This implements the new trace type icons that I've added to the `plotly-icons` lib. 

Once this https://github.com/plotly/plotly-icons/pull/16 is merged and we have a new version on npm, I'll bump it here, test it and we should be good to go!

resolves https://github.com/plotly/react-chart-editor/issues/379 https://github.com/plotly/react-chart-editor/issues/375


src/lib/strings.js
```diff
@@ -0,0 +1,49 @@
+/*eslint-disable */
+
+/**
```
VeraZab: why are we eslint disabling?

src/lib/index.js
```diff
@@ -26,6 +27,16 @@ import {
...
 import striptags from './striptags';
+import {
+  capitalize,
+  lowerCase,
+  upperCase,
+  removeNonWord,
+  camelCase,
+  pascalCase,
+} from './strings';
+
```
VeraZab: thanks :)

src/lib/strings.js
aulneau: my IDE was giving me errors related to the various replace regex like strings that are used throughout these scripts:
```
.replace(/\-/g, ' ')
.replace(/(\d)(?=(\d{1})+$)/g, '$1 ')
.replace(/\s[a-z]/g, upperCase)
.replace(/\s+/g, '')
.replace(/^[A-Z]/g, lowerCase);
```

VeraZab:
let's change plotly-icons here to `latest`: https://github.com/plotly/react-chart-editor/blob/master/package.json#L17

src/lib/strings.js
VeraZab: ok, we can keep the lint disabling

VeraZab:
💃 once plotly-icons version is changed! thanks @aulneau !!

Issue #386, opened 2018-03-07
Folds cannot be used for generic use cases without hacking around visibility


bpostlethwaite:
https://github.com/plotly/react-chart-editor/blob/master/src/components/containers/Fold.js#L29-L42

ensures that Folds will disappear if they contain children that are not plotly.js `attr` connected components. This makes it difficult for users (like the Batch Reporter) to construct their own custom panels.

VeraZab:
yeah.. well I don't see a reason why we couldn't add support for a flag (ie `show`) on custom components, that would make the Fold show if it contains a child with that flag..

Issue #385, opened 2018-03-07
if Fold children are null it throws an error


bpostlethwaite:
at this line https://github.com/plotly/react-chart-editor/blob/master/src/components/containers/Fold.js#L29

PR #389, opened 2018-03-07
Custom component in Fold fix

VeraZab:
as a fix for: #386 #385

### DIFF
--- a/src/components/containers/Fold.js
+++ b/src/components/containers/Fold.js
@@ -48,7 +48,7 @@ class Fold extends Component {
         // except for Folds of Folds, which should keep their visibility rules
         if (
           child.type.plotly_editor_traits &&
-          !child.type.plotly_editor_traits.is_fold
+          !child.type.plotly_editor_traits.foldable
         ) {
           this.foldVisible = true;
           return;
@@ -165,9 +165,5 @@ Fold.contextTypes = Object.assign(
   },
   containerConnectedContextTypes
 );
-
-Fold.plotly_editor_traits = {
-  is_fold: true,
-};
 
 export default localize(Fold);
--- a/src/components/containers/Fold.js
+++ b/src/components/containers/Fold.js
@@ -26,18 +26,30 @@ class Fold extends Component {
     this.foldVisible = false;
 
     React.Children.forEach(nextProps.children, child => {
-      if (child.props.attr) {
-        let plotProps;
-        if (child.type.supplyPlotProps) {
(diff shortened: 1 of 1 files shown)


VeraZab:
@nicolaskruchten I can't really think of anything that would break if we add this which would allow for external apps not to have to put in anything special if they want a custom component to appear in a fold. Can you?

nicolaskruchten:
This is basically similar to the `Section` pattern right, where non-`attr`-containing components count as visible by default?

VeraZab:
hmm, well, here.. even if all children are non visible.. we still want the parent to appear, like in the case with the color bar here
screen shot 2018-03-07 at 2 27 24 pm

and nesting of folds is important too whereas in Sections not really sure how that looks like. We don't want the child folds to appear if they don't contain anything:
screen shot 2018-03-07 at 2 22 59 pm

src/components/containers/Fold.js
```diff
@@ -26,18 +26,30 @@ class Fold extends Component {
     this.foldVisible = false;
 
     React.Children.forEach(nextProps.children, child => {
-      if (child.props.attr) {
-        let plotProps;
-        if (child.type.supplyPlotProps) {
-          plotProps = child.type.supplyPlotProps(child.props, nextContext);
-          if (child.type.modifyPlotProps) {
-            child.type.modifyPlotProps(child.props, nextContext, plotProps);
+      if (child) {
```
nicolaskruchten: could these two `if`s be merged?

src/components/containers/Fold.js
```diff
@@ -26,18 +26,30 @@ class Fold extends Component {
...
+            plotProps = child.type.supplyPlotProps(child.props, nextContext);
+            if (child.type.modifyPlotProps) {
+              child.type.modifyPlotProps(child.props, nextContext, plotProps);
+            }
+          } else {
+            plotProps = unpackPlotProps(child.props, nextContext);
+          }
+
+          if (plotProps.isVisible) {
+            this.foldVisible = true;
```
nicolaskruchten: where is the case handled where the child has no `attr` ?

nicolaskruchten: ah I see, below.

nicolaskruchten: ah, no, I don't see...

nicolaskruchten:
💃

VeraZab:
ok @bpostlethwaite want to try master with this fix?
should have fixed your 2 issues above!

bpostlethwaite:
Thanks @VeraZab will do!

PR #390, opened 2018-03-07
more sliders

nicolaskruchten:
Opacity set to sliders, and mousewheel should control numerics :)


VeraZab:
this is a finished pr right?
scroll the wheel, you mean the mouse wheel right? I don't have one :) can't really check it out. But nice little UX improvement, thanks 💃

nicolaskruchten:
Yep this is done, ready to merge! You can 'scroll' on a mac with two-finger drag ;)

PR #392, opened 2018-03-08
small fixes

nicolaskruchten:
* onWheel behaviour is now more limited
* fixes a Panel crash on Notes post-revision-:hocho:

### DIFF
--- a/src/components/widgets/EditableText.js
+++ b/src/components/widgets/EditableText.js
@@ -13,6 +13,7 @@ class EditableText extends Component {
     this.handleChange = this.handleChange.bind(this);
     this.handleUpdate = this.handleUpdate.bind(this);
     this.handleKeyPress = this.handleKeyPress.bind(this);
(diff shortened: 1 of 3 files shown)


src/components/widgets/EditableText.js
```diff
@@ -48,6 +49,12 @@ class EditableText extends Component {
     }
   }
 
+  handleWheel(event) {
+    if (this.props.onWheel && document.activeElement === this._ref) {
+      this.props.onWheel(event);
+    }
```
VeraZab: oh.. so you could affect the behavior even when component wasn't rendered yet..?

VeraZab:
looks good 💃

src/components/widgets/EditableText.js
nicolaskruchten: this is to prevent the onWheel to take over if the input isn't focused... the problem was that it would 'grab' the event if the mouse was just hovered over the input field, so it made just normal scrolling be annoying!

PR #395, opened 2018-03-08
fix for Fold-in-Panel-in-Fold

nicolaskruchten:


VeraZab:
💃 but can we add the test in here that I had in my previous pr:
```
it('displays fold if it contains a child that has no attr', () => {
    const wrapper = mount(
      <TestEditor {...fixtures.scatter()}>
        <Panel>
          <Fold>
            <div> ok </div>
          </Fold>
        </Panel>
      </TestEditor>
    );

    expect(wrapper.find(Fold).length).toEqual(1);
  });
```

VeraZab:
thanks its a lot better!

nicolaskruchten:
test added, will merge when CI passes

PR #396, opened 2018-03-09
dupe up colorbar and marker.colorbar

nicolaskruchten:


VeraZab:
so that's with the extra padding right?
I'm ok with the extra padding now, but there's no other UI weirdness but that right?
otherwise 💃

nicolaskruchten:
No known extra UI weirdness :) I'll add a note in #290

PR #397, opened 2018-03-09
SectionHeader

VeraZab:
@nicolaskruchten the custom section we were talking about..
I know we need tests before, but just wanted to start it out as it looks like our likely solution


bpostlethwaite:
Ha @nicolaskruchten sold me on the idea of a `<Header>` and now we are back to a logic-less `<Section>` :P

VeraZab:
well, its not really a header.. its a container.. it contains the children too..

nicolaskruchten:
💃

PR #405, opened 2018-03-13
frames support + slider styling controls

### LINKED ISSUE #391: Animations
Workspace parity around animation features (copy-paste from #325): 
- [ ] we should add the "animate by" selector unless it's super-complex. --> to be done when we're at the integration phase, requires custom plotly.js bundle and custom transform that we want to keep streambed only.

- [x] Improvement over workspace: make it such that the new editor can display/edit plots with animations that have been created in R or Python (#405 )
- [x] Improvement over workspace: some UX around positioning/styling/etc for updatemenus and sliders if they already exist in the figure, but not for creating them (#405)

VeraZab:
partially closes #391


src/EditorControls.js
```diff
@@ -30,6 +30,7 @@ class EditorControls extends Component {
       dataSourceValueRenderer: this.props.dataSourceValueRenderer,
       dataSourceOptionRenderer: this.props.dataSourceOptionRenderer,
       dictionaries: this.props.dictionaries || {},
+      frames: gd._transitionData ? gd._transitionData._frames : [],
```
VeraZab: although here I've noticed it's a little odd that my gd doesn't always get _transitionData, sometimes it's undefined..
it's a little odd because it seems to me like this key is always on gd, even if there's no frames...

like here:https://codepen.io/veraz/pen/zRRONV

nicolaskruchten:
💃 for what we've got so far! would be nice to get some of those naming bits sorted out too but that could go into a later pass

nicolaskruchten:
Also we should add a panel-link-visibility item to #290

PR #410, opened 2018-03-14
inline mocks.json for dev

nicolaskruchten:
This will make local development a little bit faster/avoid API throttles


VeraZab:
thank you 💃 !

PR #411, opened 2018-03-14
dev tools for JSON

nicolaskruchten:


dev/App.js
```diff
@@ -82,7 +83,27 @@ class App extends Component {
...
+                name={false}
+                displayDataTypes={false}
+                displayObjectSize={false}
+                indentWidth={2}
+                onAdd={({updated_src}) => this.setState(updated_src)}
+                onEdit={({updated_src}) => this.setState(updated_src)}
+                onDelete={({updated_src}) => this.setState(updated_src)}
+                src={{
+                  data: this.state.data,
+                  layout: this.state.layout,
```
VeraZab: frames

nicolaskruchten:
I didn't think it was important to edit frames/config in dev... but sure we
can add them easily enough

nicolaskruchten:
OK done, plus I got rid of the nav thing and hid it in the dev panel, to recover some real-estate. The dev app now suffers from the "tiny plot on load" problem, which is good, it'll force us to fix it for good!

VeraZab:
nice!! 💃 thanks!

PR #416, opened 2018-03-16
rangeslider/rangeselector UI

### LINKED ISSUE #68: axes type Date and Category options
Right now DefaultEditor only support Linear and Log. These could be extra options in a settings panel.

### LINKED ISSUE #330: Range Slider and Timescale Buttons panels in Axes
timescale buttons = rangeselector: https://plot.ly/javascript/reference/#layout-xaxis-rangeselector

### LINKED ISSUE #404: "Step Offset" input in StyleAxesPanel fails prop validation when axis is date
Related to #68 due to the default editor not yet supporting Date / Category types on axes.

When an axis has date values, the `Numeric` input for "Step Offset" in `StyleAxesPanel` fails prop validation for the `defaultValue` prop, due to the value being a date instead of a number.

### LINKED ISSUE #422: re-clicking autorange causes show/hide bugs
in the trace marker step/offset/min section

nicolaskruchten:
* Fixes #330 
* Fixes #422 
* Fixes #404 
* Fixes #68


src/components/fields/derived.js
```diff
@@ -78,6 +78,19 @@ export const AxisOverlayDropdown = connectToContainer(UnconnectedDropdown, {
   },
 });
 
+export const RangesliderVisible = connectToContainer(UnconnectedRadio, {
+  modifyPlotProps: (props, context, plotProps) => {
+    if (
+      !plotProps.isVisible &&
+      context.fullContainer._id &&
+      context.fullContainer._id.startsWith('x')
```
nicolaskruchten: note to self: exclude 3d traces somehow, also should provide some message for multi and y axis selection

nicolaskruchten: Vera says to use `_axisGroup` ... note to self 2: should exclude Y and All from AxesSelector where appropriate, and hide the whole panel where appropriate

nicolaskruchten: will be done as part of #424

VeraZab:
💃 💃 💃

PR #417, opened 2018-03-16
Geojson + file upload component

VeraZab:


VeraZab:
closing this for now, 
as I'm playing around with this feature I notice different mapbox errors that would need to be addressed, and since it's not the highest priority feature atm, I'm going to put this one on hold for now.
screen shot 2018-03-19 at 1 21 02 pm

Issue #333, opened 2018-02-14
Drop to upload component for images


nicolaskruchten:
Based on `react-dropzone`?

VeraZab:
yup, react-dropzone is the one we've got in streambed now

PR #418, opened 2018-03-16
Dropzone component + fixes

VeraZab:
closes: #333


VeraZab:
ready for review cc @nicolaskruchten

src/components/widgets/Dropzone.js
```diff
@@ -0,0 +1,179 @@
...
+      }
+    }
+  }
+
+  render() {
+    return (
+      <Drop
+        ref="dzone"
+        onDrop={this.onDrop}
+        disableClick={true}
```
nicolaskruchten: I would enable click, so that people can also use the file dialog instead of being required to drag.

nicolaskruchten: `/create` supports clicking on the word "upload" for this purpose, but really the whole area can be clickable and it works just as well, like I did here https://react-pivottable.js.org/

VeraZab:
screen shot 2018-03-17 at 11 09 35 am

nicolaskruchten:
Dancer!

PR #419, opened 2018-03-19
node-sass: workaround missing vendor folder

n-riesco:
After a fresh install:
```
git clone https://github.com/plotly/react-chart-editor.git
cd react-chart-editor
npm install
```

running `npm run make:lib:css` fails because of  issue https://github.com/sass/node-sass/issues/1579.

This PR adds a postinstall script to rebuild `node-sass` and fix this issue.


nicolaskruchten:
Just for context: when would you want/need to run `npm run make:lib:css` as a consumer of this library? Or is this just for clean installs of the library itself...?

PR #421, opened 2018-03-19
Scatterpolar

VeraZab:


VeraZab:
ok @nicolaskruchten would you take a look at this branch please?

src/components/containers/derived.js
```diff
@@ -18,6 +19,18 @@ const TraceTypeSection = (props, context) => {
...
+  // when TraceTypeSection is used in a connectedToLayout context
+  if (
+    fullData &&
+    fullData.some(
+      t =>
+        props.traceTypes.includes(t._fullInput.type) ||
+        fullData.some(t => props.traceTypes.includes(t.type))
+    )
+  ) {
+    return <Section {...props} />;
```
nicolaskruchten: for readability it might be nice to merge these two ifs, for example by assigning the two conditions to nicely-named functions and then `return (case1() || case2()) ? <Section> : null`...

src/components/fields/derived.js
```diff
@@ -131,12 +131,7 @@ export const AxisSide = connectToContainer(UnconnectedRadio, {
       return;
     }
 
-    plotProps.options = [
-      {label: _('Left'), value: 'left'},
-      {label: _('Right'), value: 'right'},
-      {label: _('Bottom'), value: 'bottom'},
-      {label: _('Top'), value: 'top'},
-    ];
+    plotProps.options = [{label: _('Mixed Values'), value: 'mixed values'}];
```
nicolaskruchten: this is a bit odd because clicking on this option will then reset everything to defaults no? we should disable the input in this mixed case.

src/default_panels/GraphCreatePanel.js
```diff
@@ -66,6 +68,38 @@ const GraphCreatePanel = ({localize: _}) => {
...
+          ]}
+          attr="thetaunit"
+          clearable={false}
+        />
+        <LayoutNumericSlider
+          attr="polar.sector[0]"
+          label={_('Theta Start')}
+          min={0}
+          max={360}
+          showSlider
```
nicolaskruchten: Do we need a new `LayoutNumericSlider` if we're going to use `showSlider`? Can't we just use `Numeric` with `showSlider` ?

nicolaskruchten: Or a more general `LayoutNumeric` or something?

src/default_panels/StyleAxesPanel.js
```diff
@@ -77,14 +86,17 @@ class StyleAxesPanel extends Component {
             <NumericFractionDomain label={_('End Position')} attr="domain[1]" />
           </Section>
 
-          <Section name={_('Anchor')}>
+          <TraceTypeSection
+            name={_('Anchor')}
+            traceTypes={['scatter', 'scattergl']}
```
nicolaskruchten: ternary doesn't use anchor?

src/default_panels/StyleAxesPanel.js
```diff
@@ -113,7 +125,12 @@ class StyleAxesPanel extends Component {
           </Section>
           <TraceTypeSection
             name={_('Range')}
-            traceTypes={['choropleth', 'scattergeo']}
+            traceTypes={[
```
nicolaskruchten: could we get rid of this tracetype section and merge it into the one above somehow? I think enough has changed/we've learned enough since we added this hack that we can do better now :)

src/lib/getAllAxes.js
```diff
@@ -67,9 +67,16 @@ export function axisIdToAxisName(id) {
...
+}
+
 export function getAxisTitle(axis) {
   const axisType = capitalize(axis._name.split('axis')[0]);
+  const subplotNb = getSubplotNumber(axis);
+
   return axis._input && axis._input.title
     ? striptags(`${axisType} Axis: ${axis._input.title}`)
-    : capitalize(axis._id);
+    : striptags(`${axisType} Axis ${subplotNb === 0 ? 1 : subplotNb}`);
```
nicolaskruchten: here and above the word `Axis` needs to be localized

src/default_panels/StyleAxesPanel.js
VeraZab: only cartesians

nicolaskruchten:
💃

Issue #424, opened 2018-03-20
Smart AxesSelector


nicolaskruchten:
* AxesSelector should accept a filter for axes, and hide itself if only one matches, e.g. Range Sliders when only 1 x-axis exists
* AxesSelector should not display All option for axes of different types: all 3d-cartesian are one type, all 2d-cartesian are another, all ternaries are another, angulars are another, radials are another etc.

Issue #423, opened 2018-03-20
Empty AxesFolds


nicolaskruchten:
Empty AxesFolds should either be hidden or have an empty message

PR #427, opened 2018-03-21
Smarter Axes Selector

VeraZab:
to fix: #424, #423


VeraZab:
@nicolaskruchten would you take a look?

src/lib/connectAxesToLayout.js
```diff
@@ -10,10 +10,37 @@ import {
...
 
-function computeAxesOptions(axes, _) {
-  const options = [{label: _('All'), value: 'allaxes'}];
-  for (let i = 0; i < axes.length; i++) {
-    const ax = axes[i];
+function computeAxesOptions(axes, props) {
+  const _ = props.localize;
+  let filteredAxes = axes;
+
+  if (props.name === 'Range Slider') {
```
nicolaskruchten: I would rather have all of these conditions be inlined in `src/default_panels` both in principle (because we shouldn't have this much coupling to UI details in the framework) and for localization reasons (this will all fail if "Range Slider" is in French or something :)

nicolaskruchten: i.e. we should have a new prop called `axisFilter` which is run here

VeraZab:
@nicolaskruchten 👀

nicolaskruchten:
💃 💃 💃

PR #431, opened 2018-03-23
fancier dev tools

nicolaskruchten:
* Ace editor for full control over input JSON, including copy-paste in and out.
* lazy-loading inspector for `_fullData` and `_fullLayout` to be able to drill into circular structures

### DIFF
--- a/dev/App.js
+++ b/dev/App.js
@@ -154,7 +154,7 @@ class App extends Component {
                 ]}
               />
             </Panel>
-            <Panel group="Dev" name="_fullJSON">
+            <Panel group="Dev" name="Inspector">
               <button
                 onClick={() => {
                   const gd = document.getElementById('gd') || {};
--- a/dev/App.js
+++ b/dev/App.js
@@ -110,21 +110,22 @@ class App extends Component {
         >
           <DefaultEditor>
             <Panel group="Dev" name="JSON">
-              <Select
-                clearable={false}
-                value={this.state.currentMockIndex}
-                name="mock-dropdown"
-                options={this.state.mocks.map((item, i) => ({
-                  label: item,
-                  value: i,
-                }))}
-                searchable={true}
-                searchPromptText="Search for a mock"
-                onChange={option => this.loadMock(option.value)}
-                noResultsText={'No Results'}
-                placeholder={'Search for a mock'}
-                className="mocks"
-              />
+              <div className="mocks">
+                <Select
+                  clearable={false}
+                  value={this.state.currentMockIndex}
+                  name="mock-dropdown"
+                  options={this.state.mocks.map((item, i) => ({
+                    label: item,
+                    value: i,
+                  }))}
+                  searchable={true}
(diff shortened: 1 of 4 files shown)


dev/App.js
```diff
@@ -3,9 +3,13 @@ import {hot} from 'react-hot-loader';
 import plotly from 'plotly.js/dist/plotly';
 import '../src/styles/main.scss';
 import 'react-select/dist/react-select.css';
-import ReactJson from 'react-json-view';
+import brace from 'brace'; // eslint-disable-line no-unused-vars
+import AceEditor from 'react-ace';
```
VeraZab: can we do `import 'brace'` ?

we're loading those 2 resources down below:
import 'brace/mode/json';
import 'brace/theme/textmate';

that's not all we need?

dev/App.js
```diff
@@ -59,15 +65,32 @@ class App extends Component {
...
       });
   }
 
+  updateState(data, layout, frames, currentMockIndex) {
+    this.setState({
+      data,
+      layout,
+      frames,
+      currentMockIndex,
+      full: 'hit refresh',
```
VeraZab: there's `mockIndex` in the state above, are `mockIndex` and `currentMockIndex` meant to be the same?

package.json
```diff
@@ -104,18 +103,25 @@
...
     "test:js": "jest --setupTestFrameworkScriptFile=raf/polyfill",
-    "test:lint": "eslint \"src/**/*.js\" && echo -e '\\033[0;32m'PASS'\\033[0m'",
-    "test:pretty": "prettier -l \"src/**/*.js\" && echo -e '\\033[0;32m'PASS'\\033[0m'",
-    "watch": "babel src --watch --out-dir lib --source-maps | node-sass -w src/styles/main.scss lib/react-chart-editor.css",
+    "test:lint":
+      "eslint \"src/**/*.js\" && echo -e '\\033[0;32m'PASS'\\033[0m'",
+    "test:pretty":
+      "prettier -l \"src/**/*.js\" && echo -e '\\033[0;32m'PASS'\\033[0m'",
+    "watch":
+      "babel src --watch --out-dir lib --source-maps | node-sass -w src/styles/main.scss lib/react-chart-editor.css",
```
VeraZab: I see your prettier for json is acting up now too :)

VeraZab:
well, looks good 💃 except maybe that little state typo for currentMockIndex?

dev/App.js
nicolaskruchten: I'm not sure. I followed the readme from `react-ace`

dev/App.js
nicolaskruchten: The state key is indeed `currentMockIndex`, so this is correct, but in a function above we get a new index called `mockIndex`, which we then use to overwrite.

package.json
nicolaskruchten: gah. will fix. I reset my project to try to shake the CSS bug and lost my local settings

PR #438, opened 2018-03-27
:hocho: supplyPlotProps

### LINKED ISSUE #436: get rid of supplyPlotProps
if we set a rule that only fields are `connectToContainer`ed and only containers are `connectXtoY`ed and had a visually-nil slim container we could streamline the `LayoutNumeric` stuff by just having that be in a `connectLayoutToPlot`ed slim container.

nicolaskruchten:
Fixes #436


src/default_panels/StyleTracesPanel.js
```diff
@@ -141,12 +142,12 @@ const StyleTracesPanel = ({localize: _}) => (
...
-      <LayoutNumericFractionInverse label={_('Box Width')} attr="boxgap" />
-      <LayoutNumericFraction label={_('Bar Padding')} attr="bargroupgap" />
-      <LayoutNumericFraction label={_('Box Padding')} attr="boxgroupgap" />
-    </Section>
+    <LayoutSection name={_('Size and Spacing')}>
+      <NumericFractionInverse label={_('Bar Width')} attr="bargap" />
+      <NumericFractionInverse label={_('Box Width')} attr="boxgap" />
+      <NumericFraction label={_('Bar Padding')} attr="bargroupgap" />
+      <NumericFraction label={_('Box Padding')} attr="boxgroupgap" />
+    </LayoutSection>
```
bpostlethwaite: This abstraction / API is waaaay better. It looks so freaking obvious now. All the good fixes do. Made my day!

nicolaskruchten:
@VeraZab this is now waiting on your renaming/visibility stuff and then I can 🔪 the rest of the `supplyPlotProps` machinery

src/shame.js
```diff
@@ -43,3 +43,42 @@ function clearAxisTypes(gd, traces) {
...
+        adjustedAxisIdNumber === 1 ? axis : axis + adjustedAxisIdNumber;
+
+      graphDiv.layout[
+        payload.axisAttrToAdjust + adjustedAxisIdNumber
+      ] = currentAxisLayoutProperties;
+    });
+  }
+};
+
+export const shamefullyAdjustGeo = ({layout: {geo = {}}}, {update}) => {
```
nicolaskruchten: @VeraZab this is finally how I got the geo stuff sorted. I also had to deal with the projection centers etc. Note that setting the value to `{}` essentially clears it, so this restores the defaults instead of hardcoding mappings, which is nice :)

nicolaskruchten:
Hmmm now that I look at it, it may be easier to merge my change in first... I could help you rebase :)

VeraZab:
yeah, looks like it'd be easy to rebase this one..
well, looks good to me, I'll look a little closer at the code tomorrow, but looks ready for a merge
thanks for the api fixes!

VeraZab:
💃

PR #445, opened 2018-04-01
Fixes

VeraZab:
Hey @alexcjohnson would you take a look?

### DIFF
--- a/src/components/fields/derived.js
+++ b/src/components/fields/derived.js
@@ -1,3 +1,4 @@
+import PropTypes from 'prop-types';
 import isNumeric from 'fast-isnumeric';
 import {UnconnectedDropdown} from './Dropdown';
 import {UnconnectedFlaglist} from './Flaglist';
@@ -13,7 +14,7 @@ import {
 
 export const AxisAnchorDropdown = connectToContainer(UnconnectedDropdown, {
   modifyPlotProps: (props, context, plotProps) => {
-    const {localize: _} = props;
+    const {localize: _} = context;
     let options = [];
 
     if (
@@ -94,7 +95,7 @@ export const RangesliderVisible = connectToContainer(UnconnectedRadio, {
 
 export const AxisSide = connectToContainer(UnconnectedRadio, {
   modifyPlotProps: (props, context, plotProps) => {
-    const _ = props.localize;
+    const _ = context.localize;
     if (
       context.fullContainer &&
       context.fullContainer._id &&
@@ -463,7 +464,7 @@ function computeAxesRefOptions(axes, propsAttr) {
 
 export const TextPosition = connectToContainer(UnconnectedDropdown, {
   modifyPlotProps: (props, context, plotProps) => {
-    const {localize: _} = props;
+    const {localize: _} = context;
     let options = [
       {label: _('Top Left'), value: 'top left'},
       {label: _('Top Center'), value: 'top center'},
@@ -490,7 +491,7 @@ export const TextPosition = connectToContainer(UnconnectedDropdown, {
 
 export const HoverInfo = connectToContainer(UnconnectedFlaglist, {
   modifyPlotProps: (props, context, plotProps) => {
-    const {localize: _} = props;
+    const {localize: _} = context;
     let options = [
       {label: _('X'), value: 'x'},
       {label: _('Y'), value: 'y'},
@@ -585,7 +586,7 @@ export const HoverInfo = connectToContainer(UnconnectedFlaglist, {
 
 export const FillDropdown = connectToContainer(UnconnectedDropdown, {
   modifyPlotProps: (props, context, plotProps) => {
-    const {localize: _} = props;
+    const {localize: _} = context;
 
     let options = [
       {label: _('None'), value: 'none'},
--- a/src/components/fields/derived.js
+++ b/src/components/fields/derived.js
@@ -461,14 +461,62 @@ function computeAxesRefOptions(axes, propsAttr) {
   return options;
 }
 
+export const TextPosition = connectToContainer(UnconnectedDropdown, {
+  modifyPlotProps: (props, context, plotProps) => {
+    const {localize: _} = props;
+    let options = [
+      {label: _('Top Left'), value: 'top left'},
+      {label: _('Top Center'), value: 'top center'},
+      {label: _('Top Right'), value: 'top right'},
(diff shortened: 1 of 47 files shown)


src/components/fields/derived.js
```diff
@@ -461,15 +461,63 @@ function computeAxesRefOptions(axes, propsAttr) {
...
+        'heatmap',
+        'heatmapgl',
+        'histogram2d',
+        'histogram2dcontour',
+        'contour',
+        'contourcarpet',
+        'scatter3d',
+        'surface',
+        'mesh3d',
+      ].includes(context.container.type)
```
bpostlethwaite: can't all of these come from the Plotly Schema? Doesn't the schema hold a list of all the available options for each trace? Do we need to duplicate it here?

src/default_panels/StyleTracesPanel.js
```diff
@@ -41,7 +51,14 @@ const StyleTracesPanel = ({localize: _}) => (
     />
     <NumericFraction label={_('Opacity')} attr="opacity" />
     <ColorPicker label={_('Color')} attr="color" />
-
+    <Numeric
+      showSlider
+      min={0}
+      max={1}
```
bpostlethwaite: We shouldn't need to include `min` or `max` right? That comes from the Plotly Schema and is already baked into the framework in `unpackPlotProps`. Or it was at one time.

src/default_panels/StyleTracesPanel.js
```diff
@@ -196,32 +213,19 @@ const StyleTracesPanel = ({localize: _}) => (
...
-          {label: _('Top Center'), value: 'top center'},
-          {label: _('Top Right'), value: 'top right'},
-          {label: _('Middle Left'), value: 'middle left'},
-          {label: _('Middle Center'), value: 'middle center'},
-          {label: _('Middle Right'), value: 'middle right'},
-          {label: _('Bottom Left'), value: 'bottom left'},
-          {label: _('Bottom Center'), value: 'bottom center'},
-          {label: _('Bottom Right'), value: 'bottom right'},
-        ]}
+        localize={_}
```
bpostlethwaite: Why are we passing in localize? If someone is using our API to build panels and they use `TextPosition` how will they know to pass in `localize`? No other component has this pattern as far as I am aware.

src/components/fields/derived.js
```diff
@@ -461,15 +461,63 @@ function computeAxesRefOptions(axes, propsAttr) {
...
+    let options = [
+      {label: _('Top Left'), value: 'top left'},
+      {label: _('Top Center'), value: 'top center'},
+      {label: _('Top Right'), value: 'top right'},
+      {label: _('Middle Left'), value: 'middle left'},
+      {label: _('Middle Center'), value: 'middle center'},
+      {label: _('Middle Right'), value: 'middle right'},
+      {label: _('Bottom Left'), value: 'bottom left'},
+      {label: _('Bottom Center'), value: 'bottom center'},
+      {label: _('Bottom Right'), value: 'bottom right'},
```
bpostlethwaite: Can't all these options come from the Plotly.js schema?

alexcjohnson: The values can, but not the translations.

VeraZab: I don't I've seen us use Plotly schema for options anywhere, 
Maybe it was discussed but not implemented yet :) could you give me an example?

src/default_panels/StyleTracesPanel.js
bpostlethwaite: Actually I am pretty sure it still is I think there is a test covering that behaviour

src/components/fields/derived.js
VeraZab: it doesn't come from the schema, this is not a duplicate..

src/components/fields/derived.js
bpostlethwaite: TraceSelector and Marker symbols etc did this. But @alexcjohnson is correct the translations need to be statically defined somewhere. sigh.

bpostlethwaite: @alexcjohnson it's possible that we could move the static label definitions ("defaultlabel"?) into the Schema. But that would be a fairly large undertaking. It could have some nice benefits like translated tooltips. 

Out of curiousity is that something that could be supported eventually?

src/components/fields/derived.js
alexcjohnson: Right, ideally we would pick out the `flags` from the schema for each trace type, push each into `options` here along with its translation, and then *maybe* filter based on what's actually provided in this trace - for example `text` should be available for scatter and bar traces (and it will be based on the `flags` in the schema), but it will have no effect if there's no `text` or `hovertext` field present in the trace so could be omitted.

src/components/fields/derived.js
alexcjohnson: Possible, but I'm not sure it's a good idea. Currently the schema serves as the reference for what the options are, but doesn't have anything to say about how those options are displayed in the editor. If we start moving some of the display info to the schema a) it'll be hard to know what goes where, b) it will be awkward to figure out who's responsible for translation (and when this should occur), and c) the editor will have to wait for plotly.js if it wants to make a small display change - or perhaps the context changes a little so the grammar used by plotly.js isn't quite right... point is this display information is only used in the editor so seems like it would be better defined in the editor too.

VeraZab:
@bpostlethwaite moved {localize} into context, but kept the HOC in the API.
What do you think?

src/default_panels/GraphCreatePanel.js
```diff
@@ -85,19 +84,19 @@ const GraphCreatePanel = ({localize: _}) => {
       </TraceTypeSection>
 
       <PlotlySection name={_('Axes to Use')}>
-        <AxisCreator attr="fake_attr" localize={_} />
+        <AxisCreator attr="fake_attr" />
       </PlotlySection>
 
       <PlotlySection name={_('Error Bars X')}>
-        <ErrorBars localize={_} attr="error_x" />
+        <ErrorBars attr="error_x" />
```
bpostlethwaite: :tada:

bpostlethwaite:
Wow - ha I thought you were just going a test :P 

Ok well looks like you solved that issue :tada:  Nice work!!

I think leaving in the HOC for now is ok. We can :hocho: that later if ncecessary.

PR #453, opened 2018-04-20
enable toggling the editor display

### LINKED ISSUE #448: toggle editor visibility
It would be useful to be able to hide the editor controls after a plot is setup, and just show the plot itself.  I am considering adding a `<Button/>` in a `<SingleSidebarItem/>` as the UI for this.  However, I don't see visibility options at the `<EditorControls/>` level, so it seems necessary to get deeper into the DOM in order to do this.  I wanted to check and make sure I'm not missing something obvious.  If not, I think this would be a nice feature to add.

Thank you for providing this great capability!  I am enjoying my first experiments using the React component.

bhogan-mitre:
This adds a `visible` prop to `<EditorControls/>`, and a `controlsVisible` prop to `<PlotlyEditor/>`.  The default if those props are not provided is the current behavior of displaying the editor.  If the parent  passes a value `false` in that prop, the controls are hidden but still exist in the DOM.  This enables the option to toggle the controls on and off from a parent component.  Thanks for your consideration @nicolaskruchten. 

Resolves #448

### DIFF
--- a/src/EditorControls.js
+++ b/src/EditorControls.js
@@ -22,9 +22,6 @@ class EditorControls extends Component {
       this.plotSchema = this.props.plotly.PlotSchema.get();
     }
 
-    this.state = {
-      visible: this.props.visible
-    }
   }
 
   getChildContext() {
@@ -254,7 +251,7 @@ class EditorControls extends Component {
           ' plotly-editor--theme-provider' +
           `${this.props.className ? ` ${this.props.className}` : ''}`
         }
-        style={{display: this.state.visible ? 'unset': 'none'}}
+        style={{display: (this.props.visible === undefined || this.props.visible) ? 'unset': 'none'}}
       >
         <ModalProvider>
           {this.props.graphDiv &&
--- a/src/EditorControls.js
+++ b/src/EditorControls.js
@@ -21,6 +21,10 @@ class EditorControls extends Component {
     if (this.props.plotly) {
       this.plotSchema = this.props.plotly.PlotSchema.get();
     }
+
+    this.state = {
+      visible: this.props.visible
+    }
   }
 
   getChildContext() {
@@ -250,6 +254,7 @@ class EditorControls extends Component {
           ' plotly-editor--theme-provider' +
           `${this.props.className ? ` ${this.props.className}` : ''}`
         }
+        style={{display: this.state.visible ? 'unset': 'none'}}
       >
         <ModalProvider>
           {this.props.graphDiv &&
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
We could also consider just not rendering the controls at all instead of hiding them with CSS.

bhogan-mitre:
Thanks for taking a look!  I will incorporate that and give it another try (and preserve the 'flex' display as the default, rather than 'unset').

I was thinking that hiding the controls rather than not rendering them would make it easier to bring them back up to make further plot tweaks, if necessary.  I don't know if that's a valid concern or not.

nicolaskruchten:
I would be more in favour of not-rendering if that's at all possible. It'll be more performant all around.

bhogan-mitre:
Sure, that makes sense.  I will try just not rendering.

bhogan-mitre:
I updated the PR to conditionally render the controls based on the `controlsVisible` property of `EditorControls`.  Thanks for the suggestion.  I think that will work nicely from my end.  Let me know if you have other suggestions or questions.

nicolaskruchten:
I think we'll need this to be added to `propTypes` and then you can set a default value  to skip the `undefined` check. My preference would also be for it to be called `hideControls` defaulting to `false` but that's just me :)

bhogan-mitre:
You got it!  `hideControls` property now defaults to `false`

nicolaskruchten:
This is great, thanks! Unfortunately I'm having trouble with CircleCI on this branch so the "merge" button is hidden for me. Would you be able to close and resubmit this PR and I'll merge it right away? If not, I can just make these changes directly on master if you don't care about your username being listed in the contributors :)

As a follow-up question: where are you putting the UI element that someone can use to show/hide the controls? I was thinking that post-merge I could add another prop like `showControlsToggle` which if true would render a little triangle that the user could click on to show or hide the controls. Would that be useful for you?

bhogan-mitre:
I closed and reopened, but I see there is a failing check.  Is that the CircleCI trouble you are referring to?

As for the UI, I was thinking maybe a button in the editor sidebar to hide the controls, and button in the Plotly modebar to show the controls again.  Without this merged in, I haven't tried it in my application so I can't say for sure.  I like the idea of leveraging the modebar's ability to hide itself to keep things looking clean.  But I think the toggle you mention could be useful as well.

Let me know if you'd like a separate PR to try to get this through.  I'm not sure if that's what you meant above vs. just reopening this one.

bhogan-mitre:
Okay, I think there was an issue with `prettier` before that is now resolved.  Let me know if this works for you.  Thanks!

nicolaskruchten:
This is perfect, thanks for bearing with the process overhead :)

PR #455, opened 2018-04-27
clean up trace selector

nicolaskruchten:
* Localize and rename modal to say Trace instead of Chart
* Remove HeatmapGL as per plotly.js team update
* Standardize scatter polar labelling
* Moving Specialized left so as to make it clearer that the Polar Scatter GL is an optimized version of Polar (i.e. the "X GL" are all to the right of the corresponding "X")

### DIFF
--- a/src/components/widgets/TraceTypeSelector.js
+++ b/src/components/widgets/TraceTypeSelector.js
@@ -122,8 +122,9 @@ class TraceTypeSelector extends Component {
   }
 
   render() {
+    const {localize: _} = this.context;
     return (
-      <Modal title="Select Chart Type">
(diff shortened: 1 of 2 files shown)


VeraZab:
>  the "X GL" are all to the right of the corresponding "X"

oh, that doesn't make it so there's too many basic trace types?

nicolaskruchten:
No, I just swapped the position of financial and specialized, I didn't move anything from one box to another.

VeraZab:
ok :) i'll trust it looks good 💃

nicolaskruchten:
image

nicolaskruchten:
Looks much the same :)

PR #457, opened 2018-05-02
Properly map schema, add optional tooltips

nicolaskruchten:


src/components/containers/PlotlyFold.js
```diff
@@ -104,8 +104,8 @@ Fold.propTypes = {
   canDelete: PropTypes.bool,
   children: PropTypes.node,
   className: PropTypes.string,
-  folded: PropTypes.bool.isRequired,
```
nicolaskruchten: this was giving us spurious warnings

src/lib/connectAggregationToTransform.js
```diff
@@ -27,6 +27,10 @@ export default function connectAggregationToTransform(WrappedComponent) {
 
     getChildContext() {
       return {
+        getValObject: attr =>
```
nicolaskruchten: here and in the next few files I'm just making sure each `connectXToY` properly rewraps the top-level `getValObject` calls.

VeraZab:
💃

PR #458, opened 2018-05-03
Remove key and keysrc when no dataSources

VeraZab:
@bpostlethwaite would you 💃 ?
This adjusts the dereferencing function so as to remove the parent[key] and parent[keysrc] from dereferenced object, if dataSources is an empty object.

It fixes the case in streambed where the plot keeps its data, even if the column referenced in it has been removed.

### DIFF
--- a/src/lib/dereference.js
+++ b/src/lib/dereference.js
@@ -2,7 +2,7 @@ import walkObject from './walkObject';
 
 const SRC_ATTR_PATTERN = /src$/;
 
-export default function dereference(container, dataSources) {
+export default function dereference(container, dataSources, deleteKeys) {
   const replacer = (key, parent) => {
     if (!SRC_ATTR_PATTERN.test(key)) {
       return;
@@ -12,12 +12,12 @@ export default function dereference(container, dataSources) {
     const data = dataSources[srcRef];
     const dataKey = key.replace(SRC_ATTR_PATTERN, '');
 
-    if (!Array.isArray(data)) {
-      if (Object.keys(dataSources).length === 0) {
-        delete parent[dataKey];
-        delete parent[key];
-      }
+    if (deleteKeys && !(srcRef in dataSources)) {
+      delete parent[dataKey];
+      return;
+    }
 
+    if (!Array.isArray(data)) {
       return;
     }
 
--- a/src/lib/dereference.js
+++ b/src/lib/dereference.js
@@ -2,7 +2,11 @@ import walkObject from './walkObject';
 
 const SRC_ATTR_PATTERN = /src$/;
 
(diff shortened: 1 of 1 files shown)


bpostlethwaite:
Perhaps we should lift the dereference function from editor and add it to a utility in streambed then we can extend it with impunity to fit the streambed use cases?

When will dataSources be an empty object? What happens in this case?

```javascript
dataSources = {
   a: [1,2,3]
}
data = [{xsrc: 'b', ysrc: 'a'}]
dereference(data, dataSources)
```

VeraZab:
In streambed, dataSources are an empty object when none of the grids have a used column (all columns are empty).

Yeah I think that instead of looking for an empty dataSources object, I should be looking if data is defined, if it isn't, then delete the keys.

```
    if (!Array.isArray(data)) {
      if (!data) {
        delete parent[dataKey];
        delete parent[key];
      }
      return;
    }
```

maybe even just:
```
      if (!data || !Array.isArray(data)) {
        delete parent[dataKey];
        delete parent[key];
        return;
      }
```

Would there be a situation where data is not an array but we don't want to delete the key from the parent?

bpostlethwaite:
> Would there be a situation where data is not an array but we don't want to delete the key from the parent? 
Not 100% sure. I don't think we should be too prescriptive here. What about checking for key existence instead of the data type of the dataSource value?

Hmmm if we delete the `src` reference from the figure we are losing information about what data the figure is asking for. Just because we don't have that data when dereference is called doesn't mean we won't have it at some other time. What is the case for deleting it? 

Also if we put deletion into dereference we can't pass in partial dataSources. I remember relying on that behaviour before. We could put this behaviour behind a option flag or make a new function like `syncFigure(figure, dataSources)`.

VeraZab:
> What about checking for key existence instead of the data type of the dataSource value?

yeah, that is better. We've spoken briefly with Nic about the possibility of loading in json templates of a figure, with srcs instead of values, and I think its a good idea, so true, removing srcs is maybe not what we'd like to do.

> We could put this deletion behind a option flag

I think that's reasonable, I'm going to do that. Thanks!

VeraZab:
ok @bpostlethwaite I adjusted it

src/lib/dereference.js
```diff
@@ -2,20 +2,25 @@ import walkObject from './walkObject';
 
 const SRC_ATTR_PATTERN = /src$/;
 
-export default function dereference(container, dataSources) {
+export default function dereference(container, dataSources, deleteKeys) {
```
bpostlethwaite: I would prefer this to be a config field so we can add arbitrary functionality later on without needed to mess with positional parameters
```javascript
dereference(container, dataSources, config = {})
```

PR #462, opened 2018-05-04
UI components api adjustments

VeraZab:


src/components/PanelMenuWrapper.js
```diff
@@ -10,8 +10,8 @@ class PanelsWithSidebar extends Component {
     const opts = this.computeMenuOptions(props);
 
     this.state = {
-      group: opts[0].name,
-      panel: opts[0].panels[0],
+      group: opts[1].name,
```
nicolaskruchten: this is a bit rough... :) maybe we can iterate over the children or something and grab the first one?

VeraZab:
adjusted thanks! that's why the tests were failing too :)

nicolaskruchten:
💃

PR #471, opened 2018-05-23
Adjust array src format + add transposition logic

VeraZab:


src/components/fields/DataSelector.js
```diff
@@ -33,50 +33,59 @@ export class UnconnectedDataSelector extends Component {
...
+            'carpet',
+            'contourcarpet',
+          ].includes(props.container.type)) ||
         (props.container.type === 'table' && props.attr !== 'columnorder');
     }
+
+    if (
+      this.is2D &&
+      this.fullValue &&
+      this.fullValue.length &&
```
VeraZab: guards against empty arrays

src/lib/dereference.js
```diff
@@ -7,28 +8,42 @@ export default function dereference(
...
 
-    const srcRef = parent[key];
-    const data = dataSources[srcRef];
     const dataKey = key.replace(SRC_ATTR_PATTERN, '');
+    const traceType = parent.type;
 
-    if (config.deleteKeys && !(srcRef in dataSources)) {
-      delete parent[dataKey];
-      return;
+    // making this into an array to more easily lookup 1d and 2d srcs in dataSourceOptions
```
VeraZab: it's easier to have this in an array format to lookup srcs in dataOptions, this behaviour here was broken in current dereferencing function for array srcs

nicolaskruchten:
OK so on the src-handling front, the API is a bit murky for me... Right now we expect `src`s to either be strings or arrays of strings, correct? And we want to provide a pair of functions that allow someone integrating the editor to be able to pass in some other format (i.e. encoded strings like we have in streambed) and the `joinSrc` and `splitSrc` functions are mappings between what the editor wants to process and what is externally needed. So I would rename those `toSrc` and `fromSrc` or something, to make it a bit clearer what the API is.

nicolaskruchten:
I would also have those functions be distinct props, rather than keys in an opaque-looking `customSrcHandling` object prop :)

VeraZab:
Thanks @nicolaskruchten ! Made the adjustments.

I think I like keeping the `customSrcHandling` object because I think it's maybe clearer that both `toSrc` and `fromSrc` are needed if we're going to customSrcHandling route. I added `Proptypes.shape` to make it more explicit what that object contains.

It's ready for a second look when you have a chance.

nicolaskruchten:
💃  from me on the high-level API. The tests seem good to me.

nicolaskruchten:
sorry, hit wrong button :)

Issue #470, opened 2018-05-18
[Question] Custom Attrs?


dmossie:
Hey there, I am wondering if it is possible to add custom fields with custom attrs to the react chart editor.

What I am trying to do specifically, is add a new drop down to the Trace Menu. This dropdown will influence the data options for X and Y from our database.

Example:
![untitled](https://user-images.githubusercontent.com/30665550/40261112-b381e402-5ab3-11e8-90d6-6e9a0fde729f.png)

I noticed the Dropdown and the DataSelector require an attr field. The attr field would be nice to use if it stores the data in the chart, however I do not want it to influence anything on the chart directly.

Is it possible to create custom attrs so that I can add new custom fields to the chart editor?

nicolaskruchten:
You can add custom fields and widgets, you just can't use the `attr` prop, otherwise the framework will hide them as they won't be found in `_fullData` or `_fullLayout` from plotly.js...

dmossie:
Hey @nicolaskruchten , thanks for the reply!
I find without the attr prop, the fields will not show up at all.
Example:
```
            <PlotlySection name={_('TestSection')}>
                <DataSelector label={_('TestSelector')}/>
                <Dropdown
                label={_('TestDropdown')}
                clearable={false}
                options={[
                    { label: _('Auto'), value: 'auto' },
                    { label: _('45'), value: 45 },
                    { label: _('90'), value: 90 },
                    { label: _('135'), value: 135 },
                    { label: _('180'), value: 180 },
                ]}
                />
            </PlotlySection>
```

I am assuming I can not use these components at all and that I should make my own? This might make sense as I will need an onChange function.

The reason I am asking is it would be nice to keep our components looking the same as plotlys components visually.

nicolaskruchten:
Ah, sorry, now that I look at how we've done this in the past, we've not used the exported *fields* but the *widgets* and then rewrapped them in `<Field>` but not called `connectToContainer()` on them. The resulting component has the `<Field>` styling but not the visibility rules tied to `attr`. We had at the time (3 weeks ago, at the office) discussed the idea of exporting the unconnected fields as well as the connected ones for this exact use-case and in fact we should. If you submitted a PR that did just that we'd likely approve it and re-release in a flash :)

To answer your other question: you cannot create your own `attr`s as those are defined in the [plotly.js schema](https://plot.ly/javascript/reference/)

/cc @VeraZab

dmossie:
[...]

PR #480, opened 2018-06-08
Detached DataSelector

dmossie:
Hey Plotly,

I have added a detached_fields folder. This folder can contain fields that are 'detached' or do not affect the chart. I added a DetachedDataSelector, which is a dropdown that has the same style as the usual dropdown, except it does not affect the graph.

Thank you. As discussed with @nicolaskruchten, it would be greatly appreciated if this can be contributed and re-released quickly.

### DIFF
--- a/dev/App.js
+++ b/dev/App.js
@@ -12,7 +12,7 @@ import 'brace/mode/json';
 import 'brace/theme/textmate';
 
 // https://github.com/plotly/react-chart-editor#mapbox-access-tokens
-import ACCESS_TOKENS from '../accessTokens';
+//import ACCESS_TOKENS from '../accessTokens';
 
 const dataSources = {
   ints: [1, 2, 3, 4, 5, 6], // eslint-disable-line no-magic-numbers
@@ -36,7 +36,7 @@ const dataSourceOptions = Object.keys(dataSources).map(name => ({
   label: name,
 }));
 
-const config = {mapboxAccessToken: ACCESS_TOKENS.MAPBOX, editable: true};
+const config = {/*mapboxAccessToken: ACCESS_TOKENS.MAPBOX,*/ editable: true};
 
 class App extends Component {
   constructor() {
--- a/dev/App.js
+++ b/dev/App.js
@@ -6,13 +6,13 @@ import 'react-select/dist/react-select.css';
 import brace from 'brace'; // eslint-disable-line no-unused-vars
 import AceEditor from 'react-ace';
 import Select from 'react-select';
-import PlotlyEditor, {DefaultEditor, Panel, DetachedDataSelector} from '../src';
+import PlotlyEditor, {DefaultEditor, Panel} from '../src';
 import Inspector from 'react-inspector';
 import 'brace/mode/json';
 import 'brace/theme/textmate';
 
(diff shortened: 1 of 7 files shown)


nicolaskruchten:
Thanks for the PR!

I'm a bit confused about what this is for, and the `DataSelector` naming... Can you explain a bit please?

nicolaskruchten:
Ah, I think I see, this is a response to https://github.com/plotly/react-chart-editor/issues/470#issuecomment-392625618 ?

nicolaskruchten:
If so, I think there might be a miscommunication... What I meant was that the files in https://github.com/plotly/react-chart-editor/tree/master/src/components/fields could all export the unconnected version of the field as well as the connected version, and then just thread that through the `index.js` tree. No need to duplicate all this code in a separate folder.

The other thing that confused me on the naming side of things is that `DataSelector` is not just a dropdown, it's specifically a dropdown whose contents are the available `dataSources`.

nicolaskruchten:
As I look at this a bit more closely, I see that we already do export the unconnected version of `Dropdown` (and some others... I guess we've been down this road before!) here https://github.com/plotly/react-chart-editor/blob/master/src/components/fields/Dropdown.js#L7

So you should be able to do something like 

```javascript
import {UnconnectedDropdown} from 'react-chart-editor/lib/components/Dropdown';
``` 

to get something that works almost exactly like your `DetachedDataSelector` (modulo the handling of `multi`, for example)

nicolaskruchten:
Can you confirm that the existing `UnconnectedDropdown` doesn't meet your needs first?

dmossie:
The [UnconnectedDropdown](https://github.com/plotly/react-chart-editor/blob/master/src/components/fields/Dropdown.js) looks pretty similair, but we do not want it connected to the graph (connectToContainer)

nicolaskruchten:
That's right, `UnconnectedDropdown` is not connected. This file exports both the connected (at the bottom) and unconnected versions of dropdown already.

dmossie:
Oh! That might be what we want then. Is it possible to import this from 'react-chart-editor- though? I see this in the index:
`import Dropdown from './Dropdown';`

But not the UnconnectedDropdown specifically. Should I just make a pull request with this added to the index.js files?

nicolaskruchten:
See my comment above for the relevant import statement directly from the
file.

dmossie:
Ok yes thank you - but is that a relative path import?
If so, for me that looks like this:
[import {UnconnectedDropdown} from '../../../../react-chart-editor/lib/components/Dropdown';](url)
But I get this error, and adding a symlink does not seem to work (and its a node package so should belong in node_modules).
`Module not found: You attempted to import ../../../../react-chart-editor/lib/components/Dropdown which falls outside of the project src/ directory. Relative imports outside of src/ are not supported. You can either move it inside src/, or add a symlink to it from project's node_modules/`

dmossie:
Hey Nicolas, yes I did and thank you for all your help with this. UnconnectedDropdown is what I was looking for! Perhaps I will soon create a pull request so this functionality could be exported by the package also.

nicolaskruchten:
OK great. Have a nice weekend!

PR #483, opened 2018-06-11
Added unconnected dropdown to index tree

dmossie:
I added the UnconnectedDropdown to the index tree so that it can be imported from 'react-chart-editor'


nicolaskruchten:
I'm not sure we want to do this piecemeal at this time... You can still import it right now with a full path, does this not work for you?

dmossie:
Yeah I can use the full path - but I thought this is the functionality you were looking for in our previous conversation. Is there a reason it's not a good idea?

nicolaskruchten:
It's not a bad idea in principle but I would want to implement this for *all* components in one PR, not just `Dropdown`. That's what I mean by "I'm not sure we want to do this piecemeal" :)

dmossie:
Oh ok, I could probably do that pretty quick. Would that be everything in the 'Fields' folder? Also are all of the components ready to be exported?

PR #485, opened 2018-06-12
Transforms

dmt0:
Copied transforms over to 'dev' and made the following changes there:
- issue with “Aggregations” title appearing everywhere
- name all transforms with unique name - _transform: by_
- filter out transform[]… by regex at map function
- added a few fields to 'filter'

### DIFF
--- a/src/components/containers/PanelHeader.js
+++ b/src/components/containers/PanelHeader.js
@@ -25,6 +25,7 @@ class PanelHeader extends Component {
       hasOpen,
     } = this.props;
 
+    // dropdown is styled with same styles as react-select component - see _dropdown.scss
     const icon = <PlusIcon />;
     return !children && !addAction && !allowCollapse ? null : (
       <div className="panel__header">
@@ -49,7 +50,7 @@ class PanelHeader extends Component {
           ) : null}
 
           {addAction ? (
-            <div className="panel__header__action">
+            <div className="panel__header__action dropdown-container">
               <Button
                 variant="primary"
                 className="js-add-button"
--- a/src/components/containers/PanelHeader.js
+++ b/src/components/containers/PanelHeader.js
@@ -63,9 +63,10 @@ class PanelHeader extends Component {
                 label={addAction.label}
               />
               {this.state.addPanelOpen && (
-                <ModalBox onClose={this.togglePanel} relative>
+                <div className="Select"><div className="Select-menu-outer"><div className="Select-menu">
                   {addAction.handler.map(({label, handler}) => (
-                    <p
+                    <div
+                      className="Select-option"
                       key={label}
                       onClick={() => {
                         handler(this.context);
@@ -73,9 +74,9 @@ class PanelHeader extends Component {
                       }}
                     >
                       {label}
-                    </p>
+                  </div>
                   ))}
-                </ModalBox>
+                </div></div></div>
               )}
             </div>
           ) : null}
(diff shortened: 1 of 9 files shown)


src/components/containers/PanelHeader.js
```diff
@@ -63,19 +62,24 @@ class PanelHeader extends Component {
...
-                      onClick={() => {
-                        handler(this.context);
-                        this.togglePanel();
-                      }}
-                    >
-                      {label}
-                    </p>
-                  ))}
-                </ModalBox>
+                <div className="Select">
```
nicolaskruchten: please follow BEM here :)

nicolaskruchten: Please follow BEM here :)

src/styles/components/containers/_fold.scss
```diff
@@ -158,6 +158,12 @@
       border-width: 1px 1px 1px 1px;
     }
   }
+
+  // if the panel is within a fold, allow overflow for dropdown, etc.
+  .panel {
```
nicolaskruchten: please update the existing styles rather than moving them here... rationale: this is styling Panel, so belongs in `_panel.scss`

src/components/containers/PanelHeader.js
dmt0: OK, these classes were reused from react-select component, which doesn't use BEM obviously. I think we benefit from reusing them, so that this dropdown always ends up looking like a dropdown of all combo boxes. But if BEM is more important, I could copy paste them with appropriate names...

nicolaskruchten: @VeraZab what do you think? I think it makes sense to copy-paste them that way we maintain control over their look no matter what happens when we upgrade `react-select`

VeraZab: I'd be in favor of using select classes as it makes it more obvious that it's not our own css, maybe add a comment?

And maybe that `_dropdown.scss` stylesheet could be used for this component also, we want the same overrides to apply right?

so maybe we could just add the `.dropdown` class on the enclosing div to have these style overrides apply. I think it's ok, we're dealing with a dropdown here too right

src/components/containers/TransformAccordionDev.js
```diff
@@ -0,0 +1,88 @@
...
+      this.context.container.transforms &&
+      this.context.container.transforms.map(tr => {
+        let retValue = '';
+        if (tr.groupssrc) {
+          retValue = `: ${tr.groupssrc}`;
+        } else if (tr.targetsrc) {
+          retValue = `: ${tr.targetsrc}`;
+        }
+        return retValue;
+      });
```
VeraZab: this will look weird in streambed, because like we've seen, `srcs` look like `~100: 1000`

For every `src` we should at least look in the dataSourceOptions array, and pick out the label associated with the src. In the dev app the label and value happen to be the same, but that won't always be the case:
https://github.com/plotly/react-chart-editor/blob/master/dev/App.js#L34

src/default_panels/GraphTransformsPanelDev.js
```diff
@@ -0,0 +1,102 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import {
+  Radio,
+  // TransformAccordion,
```
VeraZab: delete?

src/default_panels/GraphTransformsPanelDev.js
```diff
@@ -0,0 +1,102 @@
...
+        />
+
+        <DataSelector label={_('By')} attr="groups" />
+
+        <DataSelector label={_('Filter')} attr="target" />
+
+        <Dropdown
+          attr="operation"
+          label={_('Operation')}
+          options={[
```
VeraZab: well, I know that you know, but just make sure, 
there's more options than just those right, so i'll look for those in your next prs!

VeraZab:
:dancer: once that labelling comment is addressed

VeraZab:
great 💃

PR #486, opened 2018-06-14
Add filter transform to dev

dmt0:
Implemented filter transforms


src/default_panels/GraphTransformsPanelDev.js
```diff
@@ -71,21 +72,8 @@ const GraphTransformsPanel = (props, {localize: _}) => {
 
         <DataSelector label={_('By')} attr="groups" />
 
-        <DataSelector label={_('Filter')} attr="target" />
```
nicolaskruchten: I think you still need `target` right?

src/components/fields/FilterOperation.js
```diff
@@ -0,0 +1,196 @@
...
+      operations[key].map(o => o.value).indexOf(operator) !== -1
+    ) {
+      op = key;
+      break;
+    }
+  }
+  return op;
+};
+
+class UnconnectedFilterOperation extends Component {
```
VeraZab: would you put this component in `components/fields` like for all our other connected components :)

src/default_panels/GraphTransformsPanelDev.js
```diff
@@ -72,6 +72,7 @@ const GraphTransformsPanel = (props, {localize: _}) => {
 
         <DataSelector label={_('By')} attr="groups" />
 
+        <DataSelector label={_('Target')} attr="target" />
```
VeraZab: I would have put `By` as a label here as well

dmt0:
Basically visibility rule for DataSelector now becomes: always show, unless it's in transforms

VeraZab:
afterwards 💃

PR #493, opened 2018-06-18
Fit transform adjustments

VeraZab:


src/EditorControls.js
```diff
@@ -116,7 +116,9 @@ class EditorControls extends Component {
         if (this.props.beforeAddTrace) {
           this.props.beforeAddTrace(payload);
         }
-        graphDiv.data.push({type: 'scatter', mode: 'markers'});
```
nicolaskruchten: what's the reasoning for this change?

VeraZab: a fit is a new trace in the data array, so I need to be able to have the new trace be something other than 'scatter'

nicolaskruchten:
💃

VeraZab:
ok @nicolaskruchten this is ready for another review

src/styles/components/widgets/_numeric-input.scss
```diff
@@ -18,7 +18,6 @@
 .numeric-input__wrapper {
   line-height: 20px;
   max-width: 100%;
-  width: 100%;
```
nicolaskruchten: this causes some visual diffs in Percy that don't seem awesome...

nicolaskruchten:
You'll have to walk me through some of these changes :)

src/components/containers/TraceAccordion.js
```diff
@@ -12,11 +12,23 @@ const TraceFold = connectTraceToPlot(PlotlyFold);
 class TraceAccordion extends Component {
   render() {
     const {data = [], localize: _} = this.context;
-    const {canAdd, canGroup, children, messageIfEmptyFold} = this.props;
+    const {canAdd, canGroup, children, messageIfEmptyFold, panel} = this.props;
```
nicolaskruchten: I would recommend changing `panel` to `excludeFits` or something as a boolean, just to match `canAdd` and `canGroup`

src/EditorControls.js
nicolaskruchten: if you're actually not using this, let's revert it :)

PR #494, opened 2018-06-19
Percy integration

dmt0:
Initial Percy integration.
Note: I had to deal with legacy ref usage as they were in Percy's way. But it's generally a good idea not to use them as they will eventually get deprecated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,9 +7,9 @@
     "url": "https://github.com/plotly/react-chart-editor/issues"
   },
   "dependencies": {
+    "@plotly/draft-js-export-html": "1.2.0",
     "classnames": "^2.2.5",
     "draft-js": "^0.10.4",
-    "@plotly/draft-js-export-html": "1.2.0",
     "draft-js-import-html": "^1.2.1",
     "draft-js-utils": "^1.2.0",
     "fast-isnumeric": "^1.1.1",
@@ -28,6 +28,7 @@
     "tinycolor2": "^1.4.1"
   },
   "devDependencies": {
+    "@percy/react": "^0.4.3",
     "autoprefixer": "^7.2.3",
     "babel-cli": "^6.26.0",
     "babel-core": "^6.26.0",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -118,4 +122,4 @@
     "watch": "babel src --watch --out-dir lib --source-maps | node-sass -w src/styles/main.scss lib/react-chart-editor.css",
     "watch-test": "jest --watch"
   }
-}
+}
```
nicolaskruchten: please always put newlines at the ends of files, and don't remove them if they're already there :)

src/__percy__/panels.percy.js
```diff
@@ -0,0 +1,278 @@
...
+          <Panel group={group} name={name} />
+        </PanelMenuWrapper>
+      </TestEditor>
+    </div>
+  );
+};
+
+/* eslint-disable no-magic-numbers */
+
+percySnapshot('GraphCreatePanel2', () =>
```
nicolaskruchten: can you remove the `2` here and delete all the old snapshots from Percy as well please so we don't get `GraphCreatePanel3` and `Radio` and such?

nicolaskruchten: if you can't delete stuff that's ok, I guess it'll just be 'unchanged' forever

dmt0: Unless we do the whole "make new project, disable old one", we'll be stuck not only with unchanged forever, but also with GraphCreatePanel3, because GraphCreatePanel is already taken and it doesn't let me overwrite it for some reason - Percy bug, they're now aware of it, and will probably fix it eventually.

VeraZab:
💃 so far, and I'm ok with the removal of string `refs`, looks like React is moving away from them
https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs

nicolaskruchten:
Config file, error bars and 💃 !

PR #501, opened 2018-06-27
Add table columns if none specified

### LINKED ISSUE #499: Infer headers on tables
For `table` trace types, if nothing is provided in `headers` then this editor should just grab the data source names from the columns and stick them in the `headers`

dmt0:
Closes #499

### DIFF
--- a/src/shame.js
+++ b/src/shame.js
@@ -82,3 +82,21 @@ export const shamefullyAdjustGeo = ({layout: {geo = {}}}, {update}) => {
     update['geo.center'] = {};
   }
 };
+
+export const shamefullyAddTableColumns = (graphDiv, {traceIndexes, update}) => {
+  if (
+    update['cells.values'] &&
+    (!graphDiv.data[traceIndexes[0]].header ||
+      !graphDiv.data[traceIndexes[0]].header.valuessrc)
+  ) {
+    update['header.values'] = update['cells.valuessrc'];
+  } else if (update['header.values'] === null) {
+    update['header.values'] =
+      graphDiv.data[traceIndexes[0]].cells.valuessrc || null;
+  } else if (
+    update['cells.values'] === null &&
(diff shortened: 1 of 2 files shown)


src/shame.js
```diff
@@ -82,3 +82,21 @@ export const shamefullyAdjustGeo = ({layout: {geo = {}}}, {update}) => {
...
+    (!graphDiv.data[traceIndexes[0]].header ||
+      !graphDiv.data[traceIndexes[0]].header.valuessrc)
+  ) {
+    update['header.values'] = update['cells.valuessrc'];
+  } else if (update['header.values'] === null) {
+    update['header.values'] =
+      graphDiv.data[traceIndexes[0]].cells.valuessrc || null;
+  } else if (
+    update['cells.values'] === null &&
+    !graphDiv.data[traceIndexes[0]].header.valuessrc
```
nicolaskruchten: @VeraZab do we need to use your special `from`/`to` functions here to extract an array of data source names?

VeraZab: I think it should be ok without any adjustments, as its the data selector itself that handles the `from` and `to` src conversions, everywhere else, the editor just uses its own format, its just in that component that we care about it, because that's where the fullValue is matched up to the dataSourceOptions

nicolaskruchten: ok, great, thanks!

nicolaskruchten:
💃

src/shame.js
VeraZab: it's actually going to show very weird names in strembed, like `~:100:1000`, this does nothing to `prettify` names, its just 2 functions that allow for custom src formatting between the editor and another app. 

And those 2 functions are only used in the DataSelector, because that's the only component that has to interface between dataSourceOptions, which can be in a very custom format, and the react-chart-editor's way of handling srcs: array based for multi columns, and string based otherwise.

But these functions won't make the names of `srcs` prettier..

VeraZab:
💃

PR #502, opened 2018-06-27
Try customTraces in Dev

### LINKED ISSUE #500: POC of custom trace selector
I need to see the `dev` app modified to have a custom trace selector: I need a subset of the trace types, with customized names and no 'groups'

VeraZab:
resolves #500 
screen shot 2018-06-27 at 4 32 51 pm
cc @nicolaskruchten

### DIFF
--- a/dev/App.js
+++ b/dev/App.js
@@ -12,6 +12,7 @@ import Inspector from 'react-inspector';
 import tips from './tips';
 import 'brace/mode/json';
 import 'brace/theme/textmate';
+import {categoryLayout, traceTypes, chartCategory} from 'lib/traceTypes';
 
 // https://github.com/plotly/react-chart-editor#mapbox-access-tokens
 import ACCESS_TOKENS from '../accessTokens';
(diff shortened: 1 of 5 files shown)


VeraZab:
This should work now when configured and when not configured.
This just makes me think of the fact that we don't have anything testing the traces modal on percy as of now, as the tests were green..

dev/App.js
```diff
@@ -41,6 +42,27 @@ const dataSourceOptions = Object.keys(dataSources).map(name => ({
...
+        'scattergl',
+        'bar',
+        'histogram',
+        'table',
+        'pie',
+        'box',
+        'histogram2d'
+      ].includes(t.value)
+    )).map(t => {
+      t.category = chartCategory(_).SIMPLE
```
nicolaskruchten: let's remove this if it's not needed

src/EditorControls.js
```diff
@@ -298,6 +301,7 @@ EditorControls.propTypes = {
   plotly: PropTypes.object,
   showFieldTooltips: PropTypes.bool,
   traceTypesConfig: PropTypes.object,
+  useGLDefault: PropTypes.bool,
```
nicolaskruchten: I would have this be a full object, called `defaultInitialTrace` or something, so that we could have `bar` by default if we wanted.

nicolaskruchten:
OK so can we split this into two commits: one that's intended to be merged, and one that's just demo of how to use it? :)

src/EditorControls.js
```diff
@@ -118,7 +118,13 @@ class EditorControls extends Component {
         if (this.props.beforeAddTrace) {
           this.props.beforeAddTrace(payload);
         }
-        graphDiv.data.push({type: 'scatter', mode: 'markers'});
+
+        graphDiv.data.push(
+          this.props.useAsDefaultTrace
+            ? this.props.useAsDefaultTrace
+            : {type: 'scatter', mode: 'markers'}
```
nicolaskruchten: can this just be the default value for the prop instead?

nicolaskruchten:
💃 for the first commit once default values used

Issue #473, opened 2018-05-25
Visibility bugs in Axes panel 


nicolaskruchten:
One way to replicate is to load the `ternary-lines` mock :)

![image](https://user-images.githubusercontent.com/203523/40552546-016a4722-600e-11e8-838f-e6fabd6c5940.png)

nicolaskruchten:
OK, I poked at this for a bit and I think we can move the `fixedrange` element into the Range fold, and make the Hover Projections fold only appear on `cartesian` subplots.

dmt0:
And perhaps make _Layout_ not appear on `ternary`?

dmt0:
Making Hover Projections fold only appear on `cartesian` subplots, makes it disappear on Scatter.

PR #506, opened 2018-06-28
Axes panel visibility issues

dmt0:
Addresses #473


VeraZab:
💃

src/default_panels/StyleAxesPanel.js
```diff
@@ -480,25 +493,12 @@ class StyleAxesPanel extends Component {
...
-              {label: _('Disable'), value: true},
-            ]}
-          />
-        </AxesFold>
-
         <AxesFold
           name={_('Hover Projections')}
           axisFilter={axis =>
-            !axis._name.includes('angular') && !axis._name.includes('radial')
+            !axis._name.includes('angular') &&
```
nicolaskruchten: is it possible to have `angular` in a `cartesian` subplot? I don't think so... so we can get rid of the first two checks here no?

src/default_panels/StyleAxesPanel.js
```diff
@@ -66,7 +66,11 @@ class StyleAxesPanel extends Component {
         <AxesFold
           name={_('Layout')}
           axisFilter={axis =>
-            !axis._name.includes('radial') && !axis._name.includes('angular')
+            !(
+              axis._name.includes('radial') ||
```
nicolaskruchten: here can we not do something like `_subplot.includes('polar')` or something?

nicolaskruchten:
According to Percy we've just lost the Hover Projections fold for a cartesian set of axes... can you look into that plz?

dmt0:
Yep, because subplot is not 'cartesian', but 'xaxis' or 'yaxis'

nicolaskruchten:
Ah, ok, then that logic needs changing :)

PR #536, opened 2018-06-29
Move group and filter transforms to main

### LINKED ISSUE #444: Add Group/Filter transforms
Blocked by https://github.com/plotly/plotly.js/issues/2508 and #439

dmt0:
Closes #444


src/components/containers/TransformAccordion.js
```diff
@@ -23,7 +25,9 @@ class TransformAccordion extends Component {
       {label: _('Aggregate'), type: 'aggregate'},
     ];
 
-    if (['scatter', 'bar', 'scattergl'].indexOf(fullContainer.type) === -1) {
+    const transformableCharts = ['scatter', 'bar', 'scattergl'];
+
+    if (transformableCharts.indexOf(fullContainer.type) === -1) {
```
nicolaskruchten: I think you can use `includes` here

src/components/containers/TransformAccordion.js
```diff
@@ -32,39 +36,64 @@ class TransformAccordion extends Component {
       );
     }
 
+    const transformBy =
+      container.transforms &&
+      container.transforms.map(tr => {
+        let retValue = '';
```
nicolaskruchten: This could be `const` with a ternary operator

dmt0: retValue? I don't see how - it can have 3 values, including ''

nicolaskruchten: ok

dmt0:
I see some things are not displaying in the transforms panel. I blame Percy configuration at this point. Percy snapshot also prints a bunch of stuff into the console, such as:
`plotly-basic.js:36143 WARN: Unrecognized transform type aggregate` etc. for other transform types...
Things work [OK](http://prntscr.com/k0zb9b) when I run RCE locally.

nicolaskruchten:
Ahhh interesting! We shouldn’t be using Plotly-basic!!

dmt0:
Interesting indeed. When I use plotly or plotly-with-meta, I get 
`TypeError: window.URL.createObjectURL is not a function`
Perhaps plotly-basic is the only one that doesn't assume that there's a browser window?

nicolaskruchten:
Perhaps. Worth talking to the plotly.js guys about approaches to testing. Perhaps there are or could be private methods that would allow us to populate `_full*` without depending on browser-provided infrastructure quite so much, or perhaps they already have shim code for testing.

nicolaskruchten:
try to play with various other bundles too: with/without gl support...

src/components/containers/TransformAccordion.js
VeraZab: what is `retValue`, returnValue?
would you mind giving it a name that says more what it's used for, 
I guess this one's for the label, so `columnLabel`?

VeraZab:
For the new traces tests you added (histogram, histogram2d, ..), would you mind adding them with data (x or y values).

There's more widgets that appear sometimes when the trace is drawn.

In this case, I would have just liked seing the default trace orientation be selected.

VeraZab:
💃

Issue #509, opened 2018-06-29
deleting a trace and adding one populates editor with deleted trace


VeraZab:

nicolaskruchten:
I wonder if somehow the `useAsDefaultTrace` default value isn't getting mutated somewhere along the line by plotly.js... And the reason I don't see it in Amex is that I'm actually overwriting it?

nicolaskruchten:
I wonder if the default value for  `useAsDefaultTrace` isn't getting mutated somewhere along the line and the reason I'm not seeing it in jupyterlab is that I'm overriding it?

PR #539, opened 2018-07-03
default trace fix

VeraZab:
fixes: #509
cc @nicolaskruchten

### DIFF
--- a/src/EditorControls.js
+++ b/src/EditorControls.js
@@ -119,7 +119,9 @@ class EditorControls extends Component {
           this.props.beforeAddTrace(payload);
         }
 
-        graphDiv.data.push(this.props.useAsDefaultTrace);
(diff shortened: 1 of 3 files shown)


src/EditorControls.js
```diff
@@ -119,7 +119,13 @@ class EditorControls extends Component {
           this.props.beforeAddTrace(payload);
         }
 
-        graphDiv.data.push(this.props.useAsDefaultTrace);
+        // can't use default prop because plotly.js mutates it:
+        // https://github.com/plotly/react-chart-editor/issues/509
+        graphDiv.data.push(
+          this.props.useAsDefaultTrace
```
nicolaskruchten: I think this is still vulnerable to this problem... I would go with a generator just to be sure.

nicolaskruchten:
💃

Issue #505, opened 2018-06-28
Violin plots


nicolaskruchten:
Add support for `violin` trace types

dmt0:
Outstanding issues:

- X and Y need clearer description as to what their roles are depending on their orientation - perhaps add a note same way as in histogram
- Orientation switches automatically when you select X or Y
- `scalegroup` was not implemented
- The following are not showing up, unless they are forced in json: `bandwidth`, `span`, `box.*`,  `meanline.*`

nicolaskruchten:
Also there doesn't appear to be a way to tweak the width or color of the line?

dmt0:
Raised issue in plotly.js: https://github.com/plotly/plotly.js/issues/2834

dmt0:
Some violin attrs are fixed - updated the comment above

PR #549, opened 2018-07-05
Violin plot implementation

dmt0:
Addresses #505 
Violin plot initial implementation


dmt0:
Violin has attr "points" which does same thing as "boxpoints" for boxes. Should they be same attr?

PR #550, opened 2018-07-05
Populate transform groups when adding a transform

### LINKED ISSUE #547: Transforms default data selection
When I create a transform like aggregate or split-by, the target is auto-filled but not applied: I must clear it with the `x` and then re-select it for it to take effect.

The desired behaviour would be for it to be either auto-applied OR empty.

dmt0:
Closes #547 
Populate transforms with data correctly


nicolaskruchten:
💃

PR #551, opened 2018-07-05
boxmod, barmode, violinmode, etc

### LINKED ISSUE #543: barmode, boxmode etc
These should be implemented

dmt0:
Closes #543


nicolaskruchten:
Two things:

1) can we get box and bar mocks in Percy?
2) can you move 'whisker width' into the 'size and spacing' section and make sure it's a percentage slider?

image

src/__percy__/panels.percy.js
```diff
@@ -14,6 +14,8 @@ import './percy.css';
  * To specify which panels to test with the mock, add entry to panelsToTest, else all panels will be tested
  */
 const panelsToTest = {
+  bar: ['GraphCreatePanel', 'StyleTracesPanel'],
+  box: ['GraphCreatePanel', 'StyleTracesPanel'],
```
nicolaskruchten: how come `histogram` doesn't have `StyleTracesPanel`? can we add that plz?

dmt0:
I left "Box Mean" in a section of its own - didn't know where else to stick it. Maybe he'll find a friend later as we add more fields.

nicolaskruchten:
Somehow "whisker width" no longer appears in Percy now

nicolaskruchten:
Also, whisker width should be a numeric fraction I think.

nicolaskruchten:
💃 once those whisker issues are taken care of

nicolaskruchten:
Oh I see the problem. I'd forgotten that the section I'd asked you to move it into was layout-connected. tricksy tricksy. Maybe we should just leave it in its own little section then like it was before.

nicolaskruchten:
but just leaving it not in a section is probably fine too.

dmt0:
Yeah, I know it's totally a hack. But seems to work, so I guess we'll worry about it when it gives us trouble?

Issue #556, opened 2018-07-06
Enabling transforms for histogram, box, violin


nicolaskruchten:
Could we try to enable transforms for these three trace types? If it doesn't work, can we try to figure out why?

bhogan-mitre:
I was just about to open an issue on this.  I would suggest line plot as well as the trace types you listed.  There is nothing preventing filter transforms for histogram, box, line (and presumably violin), and that would be very useful for me.

Could we enable filtering, even if aggregate (or split) doesn't make sense for those types?

nicolaskruchten:
Line is already supported AFAIK because it's a specialization of scatter. And yes, what I'd like is all three transform types for these three trace-types.

PR #557, opened 2018-07-06
add histogram axes type

### LINKED ISSUE #555: Creating multiple histogram traces fails
I get "this is embarassing".

dmt0:
closes #555


nicolaskruchten:
Does this fix the "can't create multiple traces" issue?? that's surprising! why does it fail without these lines?

dmt0:
It does indeed.
This is where the error was thrown from:
https://github.com/plotly/react-chart-editor/blob/4f3492ca01daf59b8d86f59aea9fad342a3d7cca/src/lib/getAllAxes.js#L46

nicolaskruchten:
Presumably this makes transforms available to histograms as well right? #556

dmt0:
added `histogram2dcontour`
the rest are fine

PR #558, opened 2018-07-06
Aggregate functions

### LINKED ISSUE #508: Additional aggregation functions
Coming soon to plotly.js are additional aggregation functions `range` and `change`: https://github.com/plotly/plotly.js/pull/2764

It seems like a fairly straight-forward update to accommodate those functions here, but that we'd need to wait until plotly.js cuts a release and coordinate with that.

bhogan-mitre:
Closes #508 

This is aggregating as expected for me, but is throwing `TypeError: Cannot read property 'match' of undefined` from this line in `GraphTransformsPanel.js`:
```
      .filter(aggr => aggr.target.match(/transforms\[\d*\]\./gi) === null)
```

Any idea what else may be needed?  I don't think it's necessarily due to the new functions, as it throws an error (in the dev app) when using the `first` or `count` function as well.


bhogan-mitre:
I see this is failing the test "if data provided, but no mode is provided, chooses mode according to fullData".  I'm not sure what that has to do with the aggregation functions.  If you have any pointers please let me know!

dmt0:
It's not about the aggregate functions.
Apparently plotly.js 1.39 behaves differently.
Will take a look at it...

bhogan-mitre:
Thanks for looking into it!

etpinard:
@dmt0 the plotly.js part appears to behave the same:

image

from 

image

the editor seems to pick 'scatter' instead of 'line' from plotly.js' `mode: 'lines+markers'`

image

dmt0:
@etpinard That's weird, I tried to upgrade plotly.js and run tests (no other changes to editor) and get same behavior.

nicolaskruchten:
In general I think we'll have fewer/easier to diagnose issues if we bump plotly.js versions in a separate PR and merge that, and then add new features that depend on it. 1.39.1 should be coming out soon :)

bhogan-mitre:
That makes sense.  Should I close this one?

nicolaskruchten:
no, you can just rebase on top of master when master gets the new plotly.js :)

bhogan-mitre:
This is passing `npm test` after the update to 1.39.1, but failing `npm run test:percy` with message "PERCY_TOKEN environment variable must be set."  Let me know if I need to do anything else here.  Thanks.

nicolaskruchten:
@dmt0 can you check to see what the recommended thing to do is here with Percy? Can we run Percy on external PRs like we do with CI?

dmt0:
@bhogan-mitre We fixed our CI thing, so you can merge!
The bug with `aggr.target.match` is also fixed elsewhere.

bhogan-mitre:
Great, thanks @dmt0!  I see you've caught it up with master and we get the green check now.  Go ahead and merge when ready.  I don't think you need anything else from me, right?

nicolaskruchten:
@dmt0 only we can merge, not @bhogan-mitre ... I'll do it now. thanks for the PR!

PR #566, opened 2018-07-09
Version bump 1391

dmt0:
_plotly.js_ version bump to 1.39.1

Since _plotly.js_ no longer copies UID from `_fullData` to `data`, we can't rely on UID to identify traces and from now on have to use `fullData[i]._fullInput.index`, which will correspond to index in `data`.

### DIFF
--- a/src/EditorControls.js
+++ b/src/EditorControls.js
@@ -228,19 +228,20 @@ class EditorControls extends Component {
         break;
 
       case EDITOR_ACTIONS.DELETE_TRANSFORM:
-        if (isNumeric(payload.transformIndex)) {
-          for (let i = 0; i < graphDiv.data.length; i++) {
-            if (graphDiv.data[i].uid === payload.traceUid) {
-              graphDiv.data[i].transforms.splice(payload.transformIndex, 1);
-              if (this.props.onUpdate) {
-                this.props.onUpdate(
-                  graphDiv.data.slice(),
-                  graphDiv.layout,
-                  graphDiv._transitionData._frames
-                );
-              }
-              break;
-            }
+        if (
+          isNumeric(payload.transformIndex) &&
+          payload.traceIndex < graphDiv.data.length
+        ) {
+          graphDiv.data[payload.traceIndex].transforms.splice(
+            payload.transformIndex,
+            1
+          );
+          if (this.props.onUpdate) {
+            this.props.onUpdate(
+              graphDiv.data.slice(),
+              graphDiv.layout,
+              graphDiv._transitionData._frames
+            );
           }
         }
         break;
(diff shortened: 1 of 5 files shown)


src/EditorControls.js
```diff
@@ -228,19 +228,20 @@ class EditorControls extends Component {
...
-                  graphDiv.data.slice(),
-                  graphDiv.layout,
-                  graphDiv._transitionData._frames
-                );
-              }
-              break;
-            }
+        if (
+          isNumeric(payload.transformIndex) &&
+          payload.traceIndex < graphDiv.data.length
```
VeraZab: i'm not sure I understand this line: `payload.traceIndex < graphDiv.data.length`

before we used to loop and find the trace in data that matched our transform index. how is this replaced now with the line above?

do we not need to loop, like we used to but have the comparison be:
if `(payload.transformIndex === graphDiv._fullData[i].index)` in place of `if (graphDiv.data[i].uid === payload.traceUid)`

dmt0: I figure we looped in order to find the UID, no we go by index alone. Why did we loop in the first place, when we could just do graphDiv.data.find(...)?..

VeraZab: payload.transformIndex, indicates a _fullData index?

indexes are not 1:1, an index in data could have a different index in _fullData

dmt0: plz check the discussion in plotly_js slack:

nicolas [13:40]
we're trying to match up from `data` not `fullData`


etienne [13:41]
`fullData[i]._fullInput.index` will be the index in `gd.data`
nicolas [13:42]
ok so we might be able to go from `if (trace.uid === fullData[i]._fullInput._input.uid) {`
to `if (theIndex === fullData[i]._fullInput._input.index) {`
assuming `theIndex` is the position of `trace` in the `data` array?
etienne [13:44]
yeah, that should work

src/lib/connectTraceToPlot.js
```diff
@@ -31,7 +31,7 @@ export default function connectTraceToPlot(WrappedComponent) {
 
       let fullTrace = {};
       for (let i = 0; i < fullData.length; i++) {
-        if (trace.uid === fullData[i]._fullInput._input.uid) {
+        if (traceIndexes[0] === fullData[i]._fullInput.index) {
```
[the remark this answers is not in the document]
dmt0: There's a traceIndexes array for each trace, presumably to identify traces that you get when you split or something like that? So seems like `traceIndex[0]` always points to THE trace. Or am I misunderstanding something?

src/EditorControls.js
VeraZab: ok sure, yes them it looks like you don't need that conditional, 
thanks!

src/lib/connectTraceToPlot.js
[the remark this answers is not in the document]
VeraZab: traceIndexes can be one single trace or an array of traces.
It will be an array when we're dealing with grouped traces, like in the accordions that allow you to style multiple traces at once.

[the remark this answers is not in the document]
VeraZab: 💃

PR #578, opened 2018-07-12
Gl switch

### LINKED ISSUE #537: Trace fold icon is line in scattergl mode

### LINKED ISSUE #545: Satellite maps without mapbox tokens
When MapBox tokens are missing, trace selector should not include satellite maps.

### LINKED ISSUE #565: GL switch
For the two trace-types that have a GL-accelerated variant (`scatter`, `polarscatter`) I propose that we remove them from the trace-type picker and introduce a new special derived field that turns GL on or off.

In order to rebalance the trace-type picker visually, I would move 'distributions' to the first row and rename "WebGL" to "3D" and re-split Finance away from Specialized to get 4x2-wide blocks on the second row.

dmt0:
closes #565 
closes #537
closes #545

- add "GL Acceleration" switch
- remove *gl traces from UI
- rebalance trace selection modal
- make modal close on Esc key
- fix icon for scattergl
- don't display 'Satellite maps' when no mapbox token exists

### DIFF
--- a/src/lib/traceTypes.js
+++ b/src/lib/traceTypes.js
@@ -12,10 +12,10 @@ export const chartCategory = _ => {
       value: 'CHARTS_3D',
       label: _('3D charts'),
     },
-    // FINANCIAL: {
-    //   value: 'FINANCIAL',
-    //   label: _('Finance'),
-    // },
+    FINANCIAL: {
+      value: 'FINANCIAL',
+      label: _('Finance'),
+    },
     DISTRIBUTIONS: {
       value: 'DISTRIBUTIONS',
       label: _('Distributions'),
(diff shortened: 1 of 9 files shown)


nicolaskruchten:
Actually let's try Rendering: SVG / WebGL

nicolaskruchten:
* setting `glByDefault` without setting `makeDefaultTrace` should result in a GL trace by default.
* if in `glByDefault` mode the user switches to SVG and then clicks on another GL-available trace type, it should stay in SVG mode, and only come back if they switch to a GL-unavailable trace type first.

src/lib/traceTypes.js
```diff
@@ -30,19 +30,19 @@ export const chartCategory = _ => {
     },
     WEB_GL: {
```
nicolaskruchten: can you rename this to `3D`  plz? instead of just changing the label?

nicolaskruchten:
Just one comment in the source, and curious about why the new widget isn't showing up in Percy :)

src/lib/traceTypes.js
nicolaskruchten: `THREE_DEE`

nicolaskruchten:
💃 but please create an issue for the percy thing :)

dmt0:
https://github.com/plotly/react-chart-editor/issues/582

PR #586, opened 2018-07-16
Marker size multivalue

### LINKED ISSUE #579: Constant-vs-Variable in grouped context
For the Size widget, we need to implement the changes we discussed this morning:

image

dmt0:
closes #579

also fixes a bug where incrementing size in numericals does 6+1=61 (yay javascript!)

### DIFF
--- a/src/components/widgets/NumericInput.js
+++ b/src/components/widgets/NumericInput.js
@@ -86,9 +86,9 @@ export default class NumericInput extends Component {
     let valueUpdate;
     if (isNumeric(value)) {
       if (direction === 'increase') {
-        valueUpdate = value + step;
+        valueUpdate = parseFloat(value) + step;
       } else {
-        valueUpdate = value - step;
+        valueUpdate = parseFloat(value) - step;
       }
     } else {
       // if we are multi-valued and the user is incrementing or decrementing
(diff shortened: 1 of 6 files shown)


dmt0:
Something happened to sliders, looking at it...

dmt0:
This one's ready

src/lib/multiValues.js
```diff
@@ -17,10 +17,8 @@ function deepCopyPublic(value) {
 }
 
 function setMultiValuedContainer(intoObj, fromObj, key, config = {}) {
-  var intoVal = intoObj[key],
-    fromVal = fromObj[key];
-
-  var searchArrays = config.searchArrays;
+  const intoVal = intoObj[key];
```
nicolaskruchten: nice fixes in this file.

src/components/widgets/NumericInput.js
```diff
@@ -86,9 +86,9 @@ export default class NumericInput extends Component {
     let valueUpdate;
     if (isNumeric(value)) {
       if (direction === 'increase') {
-        valueUpdate = value + step;
+        valueUpdate = parseFloat(value) + step;
```
nicolaskruchten: man, can't believe we didn't have worse bugs before

nicolaskruchten:
So why is the slider for opacity all-blue now? In general the sliders are blue on the left, grey on the right so if the thingy is all the way to the left, it should be all-grey no? strange that it would change like this?

nicolaskruchten:
💃

PR #587, opened 2018-07-16
Sort transform

### LINKED ISSUE #584: Sort transforms
Should be relatively easy to drop in: https://github.com/plotly/plotly.js/blob/master/src/transforms/sort.js

dmt0:
closes #584 
also bumps plotly.js to 1.39.2


nicolaskruchten:
💃

Issue #517, opened 2018-06-29
colorscale vs marker color


nicolaskruchten:
* if no color src exists: don't show colorscale picker... should shamefullyAdjust so as to delete the colorscale entry from data
* if color src exists: don't show marker.color picker
* make sure grouped behaviour makes sense

nicolaskruchten:
As part of this issue, we should move the `marker.color` dataselector into Style/Traces, possibly as part of a combo field with a radio for constant vs variable, to switch between single color vs color scale.

Regarding grouped behaviour: in general I think that the multi-value version of a color picker in Style/Traces shouldn't be black with a note saying 'multiple values' but it should be a scale instead, with current value = synthetic scale assembled from current values, thereby allowing the user to play with assigning different sets of colors to groups. This particular idea makes a lot less sense in the other "grouped" context which is in `AxesFold`s in "all" mode, however!

PR #593, opened 2018-07-17
Marker color

### LINKED ISSUE #590: Constant-Variable marker color
* move Color from Create to Style
* constant mode = single colorpicker
* variable mode = data-selector
* bonus: move colorscale picker into this widget in variable mode
* 'multiple values' should be similar to the Size widget (note: if we move colorscale into this widget, then colorscale can have independently-multiple values from the data selector :)

dmt0:
closes #590 
addresses #517 

- composite field for `marker.color`
- have all colors in constants
- JS warning fixes
- css fix for field within field scenario

### DIFF
--- a/src/components/fields/MarkerColor.js
+++ b/src/components/fields/MarkerColor.js
@@ -1,0 +1,142 @@
+import Field from './Field';
+import PropTypes from 'prop-types';
+import React, {Component} from 'react';
+import {connectToContainer} from 'lib';
+import RadioBlocks from '../widgets/RadioBlocks';
+import Color from './Color';
+import DataSelector from './DataSelector';
+import Colorscale from './Colorscale';
+import {MULTI_VALUED, COLORS} from 'lib/constants';
+
+class UnconnectedMarkerColor extends Component {
+  constructor(props, context) {
+    super(props, context);
+
+    let type = null;
+    if (
+      !props.container.marker ||
+      (props.container.marker && !props.container.marker.colorsrc)
+    ) {
+      type = 'constant';
+    } else if (
+      props.container.marker &&
+      Array.isArray(props.container.marker.color) &&
+      props.fullContainer.marker &&
+      Array.isArray(props.fullContainer.marker.color)
+    ) {
+      type = 'variable';
+    }
+
+    this.state = {
+      type,
+      value: {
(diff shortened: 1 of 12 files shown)


src/components/fields/MarkerColor.js
```diff
@@ -0,0 +1,142 @@
...
+    if (
+      !props.container.marker ||
+      (props.container.marker && !props.container.marker.colorsrc)
+    ) {
+      type = 'constant';
+    } else if (
+      props.container.marker &&
+      Array.isArray(props.container.marker.color) &&
+      props.fullContainer.marker &&
+      Array.isArray(props.fullContainer.marker.color)
```
VeraZab: that's a very safe way to check :)
I guess that it's because we really want to check that it's user set, that's why we're looking in props.container.

And just to make sure, we're checking in fullContainer too, to see if the user hasn't given some crazy inputs that were not accepted by plotly.js.

ok!

dmt0: You're laughing but multivalue mechanism makes these things complicated. If you have one trace with variable color and one with constant, `container` could be either array or `MULTI_VALUE`, while `fullContainer` will be array for sure. When both traces are variable, both will be arrays.

src/components/fields/MarkerColor.js
```diff
@@ -0,0 +1,142 @@
...
+      type === 'constant'
+        ? {value: {constant: inputValue}}
+        : {value: {variable: inputValue}}
+    );
+    this.props.updatePlot(inputValue);
+  }
+
+  setColorScale(inputValue) {
+    this.setState({colorscale: inputValue});
+    this.context.updateContainer({['marker.colorscale']: inputValue});
```
VeraZab: we don't have to update colorsrc?

dmt0: Why would we? We're just changing colorscale. Am I missing something?

VeraZab: nvm, forgot how it works

VeraZab:
ok, well, looks good to me...
when it's merged, i'll do the multivalued portion of it, thanks!

VeraZab:
💃

VeraZab:
@dmt0 could you rebase please and merge :)

dmt0:
What do you mean by multivalued portion? It's done

PR #598, opened 2018-07-18
fix for DataSelector switching from 2d sources

### LINKED ISSUE #490: Switching between trace types when data sources are selected produces errors
Create a trace, select data sources, change trace type. In many cases there will be a console error.

dmt0:
closes #490

### DIFF
--- a/src/components/fields/DataSelector.js
+++ b/src/components/fields/DataSelector.js
@@ -45,7 +45,8 @@ export class UnconnectedDataSelector extends Component {
             'carpet',
             'contourcarpet',
           ].includes(props.container.type)) ||
-        (props.container.type === 'table' && props.attr !== 'columnorder');
+        (props.container.type === 'table' && props.attr !== 'columnorder') ||
+        Array.isArray(this.fullValue);
     }
   }
 


nicolaskruchten:
I'll let @VeraZab review this guy

src/components/fields/DataSelector.js
```diff
@@ -45,7 +45,8 @@ export class UnconnectedDataSelector extends Component {
             'carpet',
             'contourcarpet',
           ].includes(props.container.type)) ||
-        (props.container.type === 'table' && props.attr !== 'columnorder');
+        (props.container.type === 'table' && props.attr !== 'columnorder') ||
+        Array.isArray(this.fullValue);
```
VeraZab: well, I don't think all this.fullValues that are arrays are 2d..
2d means an array of arrays, 
so, I would check that if fullValue is an array, that all the elements in it are arrays too..

dmt0: No, this.fullValue is made of datasources, not data.
So for tables, contour Z, etc. it looks like ['ints', 'big ints'...]
Would it be different in cloud?

VeraZab: oh ok, so you're saying that when only one column can be selected, this.fullValue will not be an array? it will just be a string?

if that's the case, then ok, yes, I approuve of this logic change

dmt0: yep, it's a string like "ints"

VeraZab: ok 💃 thanks!

PR #600, opened 2018-07-19
Colorpicker adjustments

VeraZab:
new color picker
ex

demo app in react-colorscales project
screen shot 2018-07-24 at 3 34 17 pm

### DIFF
--- a/dev/App.js
+++ b/dev/App.js
@@ -10,7 +10,7 @@ import PlotlyEditor, {DefaultEditor, Panel} from '../src';
 import Inspector from 'react-inspector';
 import tips from './tips';
 import 'brace/mode/json';
-import 'brace/theme/textmate'; //
+import 'brace/theme/textmate';
 
 // https://github.com/plotly/react-chart-editor#mapbox-access-tokens
 import ACCESS_TOKENS from '../accessTokens';
--- a/dev/App.js
+++ b/dev/App.js
@@ -10,8 +10,7 @@ import PlotlyEditor, {DefaultEditor, Panel} from '../src';
 import Inspector from 'react-inspector';
 import tips from './tips';
 import 'brace/mode/json';
-import 'brace/theme/textmate';
-import {categoryLayout, traceTypes, chartCategory} from 'lib/traceTypes';
+import 'brace/theme/textmate'; //
 
 // https://github.com/plotly/react-chart-editor#mapbox-access-tokens
 import ACCESS_TOKENS from '../accessTokens';
--- a/src/components/fields/Colorscale.js
+++ b/src/components/fields/Colorscale.js
@@ -30,20 +30,11 @@ class Colorscale extends Component {
       ? fullValue.map(v => v[1])
       : null;
 
-    // removing label prop on colorscales, not enough space
-    const adjustedProps = {...this.props};
-    let fullWidth = false;
-    if (adjustedProps.label) {
-      fullWidth = true;
-      delete adjustedProps.label;
-    }
-
     return (
-      <Field {...adjustedProps}>
+      <Field {...this.props} label={false}>
         <ColorscalePicker
           selected={colorscale}
           onColorscaleChange={this.onUpdate}
-          fullWidth={fullWidth}
         />
       </Field>
     );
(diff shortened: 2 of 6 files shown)


dev/App.js
```diff
@@ -10,8 +10,7 @@ import PlotlyEditor, {DefaultEditor, Panel} from '../src';
 import Inspector from 'react-inspector';
 import tips from './tips';
 import 'brace/mode/json';
-import 'brace/theme/textmate';
-import {categoryLayout, traceTypes, chartCategory} from 'lib/traceTypes';
+import 'brace/theme/textmate'; //
```
dmt0: empty comment...

src/components/fields/Colorscale.js
```diff
@@ -30,11 +30,20 @@ class Colorscale extends Component {
...
+    const adjustedProps = {...this.props};
+    let fullWidth = false;
+    if (adjustedProps.label) {
+      fullWidth = true;
+      delete adjustedProps.label;
+    }
+
     return (
-      <Field {...this.props}>
+      <Field {...adjustedProps}>
```
dmt0: Could you just do `<Field {...this.props} label={false}>`?
I think if a props is defined twice, only the last one counts, so you can overwrite with an empty.

VeraZab: that's nice, didn't think of that one thanks!

VeraZab:
Hey @jackparmer I just updated the react-colorscales project page, to include all api changes:
https://github.com/plotly/react-colorscales

I changed the look of the colorPicker to adapt to small UIs, but I have only extended the api, so it should be ok for the projects using it.

I've put up a few screenshots above, and deployed to gh-pages:
https://plotly.github.io/react-colorscales/

Let me know if you'd like any adjustments :)

jackparmer:
Hey, cool! I love it! 

Only nit that caught me eye is `Cmocean` should be all lowercase (`cmocean`). That seems to be how it's presented in the literature: https://matplotlib.org/cmocean/

image

nicolaskruchten:
I'm a bit confused by the UX of the new picker here...

image

There's no clear indication of what the active scale is, or of how to "close" the picker, or why "reset" is there... I think I like this layout but I preferred the previous setup where the active scale is always present above, and the panel opens when you're fiddling with it and it updates the activate scale, and then you can close the panel when you're done fiddling.

VeraZab:
ex

ok I corrected titles, they're small caps now, just for a consistent look. 
There's also the 'scale' word in the label to make what we're selecting clearer.

@nicolaskruchten , I added back the toggle that you liked and on another UX pass, as you said, we could think of a more discoverable way to make these controls open/close.

ex

src/components/widgets/ColorscalePicker.js
```diff
@@ -1,41 +1,95 @@
...
+
   render() {
-    const {selected, onColorscaleChange} = this.props;
+    const {onColorscaleChange, selected} = this.props;
+    const {selectedColorscaleType, showColorscalePicker} = this.state;
+    const description =
+      COLOR_PICKER_CONSTANTS.COLORSCALE_DESCRIPTIONS[selectedColorscaleType];
+    const colorscaleOptions = COLOR_PICKER_CONSTANTS.COLORSCALE_TYPES.map(
+      type => ({
+        label: type + ' scale',
```
nicolaskruchten: actually now that I look at it, could we make this `scales` with an `s` please? :)

src/components/widgets/ColorscalePicker.js
```diff
@@ -1,41 +1,95 @@
...
+import React, {Component, Fragment} from 'react';
 
 class Scale extends Component {
-  constructor(props) {
-    super(props);
+  constructor() {
+    super();
+
     this.state = {
+      selectedColorscaleType: 'sequential',
```
nicolaskruchten: note: in a group of traces, this default makes less sense, and should be overrideable to `categorical`

VeraZab: yup, will look at how to select the initial state better in a subsequent pr..
but the structure is there to add it easily now :)

VeraZab:
ok @nicolaskruchten did the latest color/swatch number modifications
screen shot 2018-07-24 at 5 50 53 pm

nicolaskruchten:
💃

Issue #575, opened 2018-07-11
Subplot axis names behave weirdly


nicolaskruchten:

nicolaskruchten:
@dmt0 anything left to do here or can we close this guy out?

PR #601, opened 2018-07-19
Axis names fix

### LINKED ISSUE #479: make the option to disable overlaying more discoverable
This is a suggestion after trying to explain to a Falcon user how to create a subplot (see https://github.com/plotly/falcon-sql-client/issues/449).

The step I found hard to discover was how to disable overlaying:

screenshot from 2018-06-07 10-49-58

How do you think about having the drop-down menu show `disabled` as an option?

screenshot from 2018-06-07 11-00-29

dmt0:
addresses #575 
closes #479 

- refactored axes component naming
- correct axes name during creation
- axes names in axes panel
- add None option for Overlay


nicolaskruchten:
@VeraZab can you review this as well please, given that you spent so much time working on this initially? :)

src/components/fields/AxesSelector.js
```diff
@@ -16,36 +16,45 @@ class AxesSelector extends Component {
...
-      return (
-        <Field {...this.props} label={_('Axis to Style')}>
-          <Dropdown
-            options={axesOptions.map(option => {
-              if (option.value !== 'allaxes') {
-                return {
+    const multipleSublots =
+      fullLayout &&
+      fullLayout._subplots &&
+      Object.values(fullLayout._subplots).some(s => s.length > 1);
```
VeraZab: hmm, fullLayout._subplots is an object right
that contains subplotTypes as keys, and arrays as values, and we want to check that those arrays are of length greater than 1.

so isn't it more like:
`Object.values(fullLayout._subplots).some(s => fullLayout[s].length > 1);`

VeraZab:
other than the above, looks good to me!

src/components/fields/AxesSelector.js
dmt0: Vera, it's Object.values(), not Object.keys()

VeraZab:
💃

nicolaskruchten:
can you force-push here to re-trigger circle and percy plz?

dmt0:
CircleCI is still down as of 6 minutes ago

dmt0:
I'm thinking Overlay should have an explicit option for "none" - it took me a while to figure out that I have to clear it to spread out the plots.

nicolaskruchten:
Agreed, please add 'none'. This will close #479

dmt0:
err, am I still passing review after all these commits? :)

nicolaskruchten:
@VeraZab please re-review if needed! I think the behaviour is fine for me. The only known-weird thing left outstanding with the secondary y-axis thing is https://github.com/plotly/plotly.js/issues/2828

VeraZab:
ok rechecked looks good 💃

Issue #524, opened 2018-06-29
systematically compare workspace and editor attr coverage


nicolaskruchten:
* @dmt0 and @nicolaskruchten will go trace type by trace type and compare workspace and editor
* @VeraZab will do it code-wise

dmt0:
So far I went through Scatter, Bar, Line, Area, and Pie, it resulted in https://github.com/plotly/react-chart-editor/pull/561. Will also raise a few issues that were not included.

dmt0:
Went through all Geo/Mapbox, 3d, Ternary, that resulted in #589 and #660

dmt0:
Went through Mapbox

PR #605, opened 2018-07-23
3d trace types - feature parity

### LINKED ISSUE #516: heatmap colorscale: zmin/zmax don't hide when zauto is true
plotly.js bug?

### LINKED ISSUE #520: colorbar bgcolor doesn't work (or in workspace)
plotly.js bug?

### LINKED ISSUE #606: Standardize colorscale attrs
Right now we support some nice color-related attrs in Style/Trace for the surface/heatmap traces that should also appear in scatter and other non-covering traces, like `showscale`, `reversescale`, `cmin`, `cauto` etc. They're mapped to `marker.X` in the latter case, and for the various scatter options, they also exist for `line.X`.

Also, there's some weirdness in Heatmap traces around `zauto` which should be resolved similarly...

Surface (`c*`):

image

Heatmap (`z*`):

image

Scatter (`marker.c* missing`):

image

dmt0:
**General**
- Menu order: Traces, Layout, Axes, Legend, Color Bars, Notes, Shapes, Images, Sliders, Menus
- Place Violin next to Box

**3D scatter**

StyleTrace
 - can’t select data source for text
 - on hover - ‘text’ missing
 - Line type broken when multiple traces - can’t get color

Layout
 - aspect ratio missing - manual mode broken in cloud sometimes - setting xyz brings back to auto

Axes
 - Empty sections: Layout

Color Bars
 - Tick color -> Color
 - Borders - Background Color broken - also broken in cloud - plotly returns rgba instead of rgb, when you change it to rgb in json, ColorPicker works  "rgba(245, 117, 117, 0)"

Images
 - Relative position to grid missing
	

**3D Line**

StyleTrace
 - rename Line Color -> Color

Axes
 - Empty sections: Layout

**3D Surface**

StyleTrace
 - Colorscale Range missing - auto/custom, min, max

Layout
 - aspect ratio missing

Color Bars
 - rename Vertical padding -> Padding, same for Horizontal

**3D Mesh**

StyleTrace
 - Show Contours - should hide color and width when “hide”

Layout
 - aspect ratio missing
 - Empty sections: Layout

Color Bars
 - Horizontal Positioning - rename horizontal position -> position

**Ternary**
- hover actions - replace flaglist with dorpdown, since you can't deselect both and it behaves funny

Closes #516 
Closes #520 
Closes #606 
Addresses #524


src/components/fields/VisibilitySelect.js
```diff
@@ -0,0 +1,80 @@
+import React, {Fragment, Component} from 'react';
+import PropTypes from 'prop-types';
+import {connectToContainer} from 'lib';
+import Field from './Field';
+import Radio from './Radio';
+import Dropdown from './Dropdown';
+import nestedProperty from 'plotly.js/src/lib/nested_property';
+
+class UnconnectedVisibilitySelect extends Component {
```
nicolaskruchten: how does this guy handle multiple values ?

src/default_panels/StyleTracesPanel.js
```diff
@@ -337,7 +337,13 @@ const StyleTracesPanel = (props, {localize: _}) => (
     </PlotlySection>
     <TraceTypeSection
       name={_('Text')}
-      traceTypes={['scatter', 'scatterpolar', 'scatterpolargl', 'pie']}
+      traceTypes={[
+        'scatter',
```
nicolaskruchten: scattergl also? generally they go in pairs :)

src/default_panels/StyleTracesPanel.js
```diff
@@ -593,13 +593,15 @@ const StyleTracesPanel = (props, {localize: _}) => (
         'scatterpolargl',
       ]}
     >
-      <Flaglist
+      <Dropdown
```
nicolaskruchten: why this change?

dmt0: Because there are only three options - you cannot deselect both, I found it kinda funny. Subjective really.

nicolaskruchten:
Notes:

* gl3d_bunny-hull.json mock crashes Style/Traces panel
* (actually many of the gl3d mocks crash that panel)
* The `aspectratio` stuff only applies to the first `scene` in this case (i.e. it will be weird if there are two scenes), so it'll have to move to the new Subplots pane... I'm tempted to remove it in this PR and add it in when we do subplots. IIRC there was some stuff like this when we last looked at `polar` as well, so when we do the subplots panel we'll have to sweep through and find all the places where we mapped an attr to `polar.*` or `geo.*` and move those to the applicable subplot fold I guess :)

src/default_panels/StyleTracesPanel.js
```diff
@@ -241,6 +241,22 @@ const StyleTracesPanel = (props, {localize: _}) => (
       <SymbolSelector label={_('Symbol')} attr="marker.symbol" />
       <Numeric label={_('Border Width')} attr="marker.line.width" />
       <ColorPicker label={_('Border Color')} attr="marker.line.color" />
+      <Radio
```
nicolaskruchten: can you plz remove these from `line` given that we don't allow line colorscales atm :)

src/components/fields/MarkerColor.js
```diff
@@ -93,35 +96,72 @@ class UnconnectedMarkerColor extends Component {
...
-            {container.marker &&
-            container.marker.colorscale === MULTI_VALUED ? null : (
-              <Colorscale
-                suppressMultiValuedMessage
-                attr="marker.colorscale"
-                updatePlot={this.setColorScale}
-                colorscale={colorscale}
-              />
-            )}
+            <Radio
```
nicolaskruchten: from the Percy screenshot it looks like multi-value is odd with these...

nicolaskruchten: well, I could be convinced that the percy screenshot is reasonable, let's talk about it

nicolaskruchten:
💃 ! I think this also closes #520 and #606

PR #609, opened 2018-07-24
visibility nesting fix

### LINKED ISSUE #608: Atlas map regression
It used to be possible to change the projection of an Atlas Map/`scattergeo` from Create/Trace (I randomly checked an old commit 5f65565548c30dc74de2ab508b800797c979b614 and it works (note: I tried that commit with the then-contemporary version of plotly.js and then master with that same version, so the issue is most likely in the editor and not in plotly.js I think). Now you have to specify `locations` for those to show up, but that breaks the image because you can't specify points in lat/lon.

So two things:

1) for Altas Maps, it's either Locations OR lat/lon and the UI should reflect that by never showing both
2) the Options section should always be present (we should figure out what changed!)

image

nicolaskruchten:
Closes #608 🙈

### DIFF
--- a/src/default_panels/GraphCreatePanel.js
+++ b/src/default_panels/GraphCreatePanel.js
@@ -26,11 +26,11 @@ const GraphCreatePanel = (props, {localize: _}) => {
       <TraceSelector label={_('Type')} attr="type" show />
 
       <PlotlySection name={_('Data')}>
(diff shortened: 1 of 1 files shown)


dmt0:
💃

VeraZab:
💃

Issue #580, opened 2018-07-13
Subplots for all trace types


nicolaskruchten:
In an ideal world, users should be able to create and manipulate all subplot variants that plotly.js can create. Certainly it should be possible to create multiple `pie` subplots for example.

nicolaskruchten:
Types of subplots:
* 2d
* 3d
* geo
* pie
* polar
* ternary
* table/parcoords/splom??

Examples of how to subplot various combos of these:
* geo_africa-insets.json (multiple geo)
* polar_subplots.json (multiple polar)
* https://plot.ly/~nicolaskruchten/144/#code (geo + 3d + 2d)
* https://plot.ly/python/pie-charts/ (has dual-pies)
* ternary???

nicolaskruchten:
Should we consider a new "Subplots" panel which somehow figures out all the subplots and allows us to position them with vertical/horizontal sliders?

nicolaskruchten:
Step 1: let's collect in a comment in this issue information about where in the figure the subplot domain information lives for all the different types of subplots listed above :)

dmt0:
So it seems domain can be either in layout or in data and it can be either an array of two values or an object:

layout.axisname.domain = [0.65, 0.95]
xaxis, yaxis

layout.axisname.domain = {“y”: [0, 0.9], "x": [0, 0.55]}
geo, polar, scene, ternary

data[x].domain = {"x": [0, 0.48]}
parcoords, pie, table

dmt0:
Having an extra panel for this would allow us to do things without magic and keep things close to ploty.js.

But with a bit of magic... we could make a radio button in the GraphCreatePanel with choices like "Merged" | "Separate subplot" | "Move to the side within same subplot" (with better wording of course), and once that radio is selected, the new axes are created automatically as needed and sliders appear that allow to move your subplot around. That would replace current AxesSelector.

nicolaskruchten:
OK, so here's the spec for how to handle this issue:

* We add a new Style/Subplots panel, which renders as a Subplot Accordion, which has a fold per subplot. 
* The list of subplots is defined to be the union of all (xaxis-yaxis pairs which share a trace, scenes, polars, ternarys, geos, pie-traces, table-traces). 
* Each fold in this accordion has (minimally) 4 sliders: horizontal/vertical min/max domain. (we can turn these into two-knob sliders later)
[...]

PR #610, opened 2018-07-26
subplots panel initial

dmt0:
Subplots panel - in dev

Known issues / TODO:
- rectangle needs to be positioned correctly - no making it 3px doesn't work
- DualNumeric is a hack - needs to be done right
- possibly merge AxesCreator and SubplotCreator - at least partially - lots of ways to improve the code - some utility functions for figuring out subplot names would be useful - raised https://github.com/plotly/react-chart-editor/issues/631
- move geo.*, scene.* stuff into subplot panel OR create some SubpotConnectedSections in GraphCreatePanel

addresses #580


nicolaskruchten:
Dude this is sweet!

dmt0:
:)
I'm thinking, shouldn't we just move all that into GraphCreatePanel into the trace folds? The only drawback I see is that if two traces are on the same subplot, their controls will be mirroring each other.

nicolaskruchten:
So cool! non-cartesian appears broken, and dragging is broken in snap mode, but so cool!

nicolaskruchten:
In `AxesCreator` you'll need to update the text to say that you can edit the subplots in the right panel. now it says "axes"

nicolaskruchten:
image

src/components/fields/derived.js
```diff
@@ -17,22 +17,14 @@ export const AxisAnchorDropdown = connectToContainer(UnconnectedDropdown, {
     const {localize: _} = context;
     let options = [];
 
-    if (
-      plotProps.fullContainer &&
-      plotProps.fullContainer._subplot &&
-      plotProps.fullContainer._subplot.includes('xaxis')
-    ) {
+    if (props.attr.startsWith('x')) {
```
VeraZab: this seems a little fragile, is there a way to make it more precise?
what if there's other axes related attributes that start with x or y..

package.json
```diff
@@ -23,8 +23,10 @@
     "react-dropzone": "^4.2.9",
     "react-plotly.js": "^2.2.0",
     "react-rangeslider": "^2.2.0",
+    "react-resizable-rotatable-draggable": "^0.1.8",
     "react-select": "^1.2.0",
     "react-tabs": "^2.2.1",
+    "styled-components": "^3.3.3",
```
VeraZab: are we actually using it?

src/components/fields/derived.js
dmt0: good call, i'll make it more precise
but generally, this guys get attr supplied to them

package.json
nicolaskruchten: it's a peer dep of the draggy thing.

src/components/containers/SubplotAccordion.js
```diff
@@ -0,0 +1,129 @@
...
+            xaxis={d.xaxis}
+            yaxis={d.yaxis}
+            name={`${d.xaxisName} ${d.yaxisName}`}
+          >
+            {children}
+          </CartesianSubplotFold>
+        ))
+    );
+
+    Object.keys(layout).forEach(layoutKey => {
```
VeraZab: I found it confusing here as to why we were looping through layout, not fullLayout.
And why we were using data.findIndex instead of maybe filter to find all indexes that could meet the condition.

If you could maybe comment above on those 2 questions, I think it may help in a few months time, when revisiting that code.

src/components/fields/AxesCreator.js
```diff
@@ -21,62 +22,61 @@ class UnconnectedAxisCreator extends Component {
...
+      },
+    });
   }
 
-  recalcAxes(update) {
+  updateAxis(update) {
     const currentAxisId = this.props.fullContainer[this.props.attr];
+    const axesToBeGarbageCollected = [];
 
     // When we select another axis, make sure no unused axes are left
```
VeraZab: comment no longer valid

src/components/fields/RectanglePositioner.js
```diff
@@ -0,0 +1,194 @@
...
+          }}
+        >
+          {Array(gridRes * gridRes)
+            .fill(0)
+            .map((v, i) => (
+              <div
+                key={i}
+                className="rect-grid"
+                style={{
+                  width: fieldWidthPx / gridRes - 1,
```
VeraZab: could we move this to a style sheet?

nicolaskruchten: not really, if you see the math we're doing here, it doesn't really map well onto a stylesheet

nicolaskruchten: I suppose we could create classes for the various grid-box-types but there are 9 of them (4 corners, 4 faces, inside) which would be messier than just building them up from stratch like this IMO. Although with faaaancy CSS3 selectors you might be able to do it ;)

VeraZab:
awesome job guys !! 👍 
If @dmt0 you could just leave me a comment in SubplotAccordion of what the json looks like in a situation where we have 2 subplots of either ['geo', 'mapbox', 'polar', 'gl3d', 'ternary'], I'm 💃 on this pr.

nicolaskruchten:
💃

PR #611, opened 2018-07-26
Color multiple traces

VeraZab:
Enables coloring multiple traces at once, like on cloud, and includes #10053 improvements.
ex


nicolaskruchten:
As per conversation, need to look at Multiple Values handling.

VeraZab:
ex

nicolaskruchten:
💃 !

PR #617, opened 2018-07-30
Categorical scale fixes

### LINKED ISSUE #614: Categorical color scales should cycle
Right now it looks like generating colors from a categorical scale *interpolates* between the colors just like it interpolates for sequential or diverging scales. We actually want it to *cycle* through the colors in the scale.

Example: for a three-color categorical scale ABC, here is what happens today as I request more colors, vs what we would want to see (where AB means 'some color between A and B'):
 
| today | desired |
| --- | --- |
| A  |  A |
| A C  |  A B |
| A B C  |  A B C |
| A AB BC C  |  A B C A |
| A AB B BC C  |  A B C A B |

VeraZab:
Closes #614

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-chart-editor",
   "description": "plotly.js chart editor react component UI",
-  "version": "0.22.0",
+  "version": "0.22.1",
   "author": "Plotly, Inc.",
   "bugs": {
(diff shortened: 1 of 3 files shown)


nicolaskruchten:
This is meant to close #614 right?

VeraZab:
yes

nicolaskruchten:
💃

PR #620, opened 2018-07-31
Color pie traces with colorscale

### LINKED ISSUE #538: Can't change colors of pie chart slices
I might be missing something in the [demo app](https://plotly.github.io/react-chart-editor/), but it doesn't seem that one can change the pie chart slice colors:

image

image

VeraZab:
Closes #538

### DIFF
--- a/src/components/fields/Colorscale.js
+++ b/src/components/fields/Colorscale.js
@@ -13,10 +13,10 @@ class Colorscale extends Component {
   onUpdate(colorscale, colorscaleType) {
     if (Array.isArray(colorscale)) {
       if (this.context.container.type === 'pie') {
-        const pieSlices = this.context.container.labels.length;
+        const numPieSlices = this.context.container.labels.length;
         const adjustedColorscale = adjustColorscale(
           colorscale,
-          pieSlices,
+          numPieSlices,
           colorscaleType,
           {repeat: true}
         );
--- a/src/components/fields/Colorscale.js
+++ b/src/components/fields/Colorscale.js
@@ -2,7 +2,7 @@ import ColorscalePicker from '../widgets/ColorscalePicker';
 import Field from './Field';
 import PropTypes from 'prop-types';
 import React, {Component} from 'react';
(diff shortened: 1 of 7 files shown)


src/lib/index.js
```diff
@@ -175,7 +176,33 @@ function maybeAdjustSrc(src, srcAttributePath, traceType, config) {
...
+  config
+) {
+  if (config.repeat) {
+    const repetitions = Math.ceil(numberOfNeededColors / colorscale.length);
+    const newArray = new Array(repetitions).fill(colorscale);
+    return newArray
+      .reduce((a, b) => {
+        return a.concat(b);
+      }, [])
+      .slice(0, numberOfNeededColors);
```
dmt0: What do you need that newArray for?

```
    return new Array(repetitions).fill(colorscale
      .reduce((a, b) => a.concat(b), [])
      .slice(0, numberOfNeededColors);
```

src/components/fields/Colorscale.js
```diff
@@ -12,6 +12,18 @@ class Colorscale extends Component {
 
   onUpdate(colorscale, colorscaleType) {
     if (Array.isArray(colorscale)) {
+      if (this.context.container.type === 'pie') {
+        const pieSlices = this.context.container.labels.length;
```
dmt0: it's more like numPieSlices?

src/lib/index.js
VeraZab: its just my way of generating an array of colors that repeat themselves enough times to cover all the traces that need to be colored..

dmt0:
Only looked at code.
Just those two minor things, otherwise 💃!

src/lib/index.js
dmt0: yeah, I figured, just saying could be written sorter without extra consts

nicolaskruchten:
From conversation: we should generate N colors where N is the length of calcdata, not labels.

nicolaskruchten:
💃 !

PR #625, opened 2018-08-02
ColorPicker => MultiColorPicker

VeraZab:


nicolaskruchten:
Behaviour looks good, 💃 from me subject to @dmt0 's code review

src/components/widgets/ColorPicker.js
```diff
@@ -152,7 +152,7 @@ class ColorPicker extends Component {
 
 ColorPicker.propTypes = {
   onColorChange: PropTypes.func.isRequired,
-  selectedColor: PropTypes.string,
+  selectedColor: PropTypes.any,
```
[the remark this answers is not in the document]
dmt0: There's a handy construct for this:
`PropTypes.oneOfType([PropTypes.string, PropTypes.number])`

src/components/fields/MultiColorPicker.js
```diff
@@ -0,0 +1,166 @@
...
+        props.fullValue.every(v => v[1] === props.fullValue[0][1])
+          ? 'single'
+          : 'multiple',
+    };
+    this.setColor = this.setColor.bind(this);
+    this.setColors = this.setColors.bind(this);
+  }
+
+  setColor(color) {
+    this.props.updatePlot(color);
```
dmt0: why not just pass updatePlot as a prop?

src/components/fields/MultiColorPicker.js
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  render() {
+    const _ = this.context.localize;
+    const constantOptions = [
+      {label: _('Single'), value: 'single'},
+      {label: _('Multiple'), value: 'multiple'},
+    ];
+    const selectedConstantColorOption = this.props.parentState
+      ? this.props.parentState.selectedConstantColorOption
```
dmt0: parentState only has one variable, why not just have parentSelectedConstantColorOptions?

VeraZab: I'm using parent state to pass MarkerColor state, it has more than one variable.

src/components/fields/MultiColorPicker.js
VeraZab: It is passed, from the connect wrapper, I just prefer having a separate function, 
one setColor, one setColors, that's all

src/components/fields/MultiColorPicker.js
dmt0: I'm not seeing other variables

PR #630, opened 2018-08-03
Style splits

VeraZab:
Closes: #528

Here are the changes that were introduced with this pr:
- Folds are created based on the number of traces in fullData if the `useFullData` flag is used on a TraceAccordion: this is done in the StyleTracesPanel

- When the useFullData flag is used, a new prop `fullDataArrayPosition` is passed to the Fold, which indicates the index/indexes in the fullData array, that Fold represents (depending on whether the fold represents a single trace or a group of traces).

- If we're using fullData in the TraceAccordion, the format of `traceIndexes` changes as well, in the case that we have split traces. Because split traces all refer to a single trace in the data array, in the individual style tabs, we will have more than one Fold that will have the same traceIndex. Also in the grouped tab, we could have the same index repeated many times. 
Ex:
grouped scattered traces fold, could have: 
traceIndexes: [0, 0, 0, 1],
fullDataArrayPosition: [0, 1, 2, 3]

- Styling split traces, writes to their data[x].transforms[y].styles[z].value object in the individual panel AND in the grouped trace panel. In the grouped case, it wasn't enough to just write the style to the parent trace because if a style had been already present in the transforms[y].styles[z].value object, then the parent could not override it.

### DIFF
--- a/src/components/containers/TraceAccordion.js
+++ b/src/components/containers/TraceAccordion.js
@@ -137,7 +137,7 @@ class TraceAccordion extends Component {
       if (this.filteredTraces.length === 1) {
         return (
           <TraceRequiredPanel>
-            <PlotlyPanel>{this.renderUngroupedTraceFolds()}</PlotlyPanel>
+            {this.renderUngroupedTraceFolds()}
           </TraceRequiredPanel>
         );
       }
--- a/src/components/containers/TraceAccordion.js
+++ b/src/components/containers/TraceAccordion.js
@@ -42,7 +42,7 @@ class TraceAccordion extends Component {
     const fullDataArrayPositionsByTraceType = {};
 
     this.filteredTraces.forEach((trace, index) => {
-      const traceType = plotlyTraceToCustomTrace(trace.type);
+      const traceType = plotlyTraceToCustomTrace(trace);
       if (!dataArrayPositionsByTraceType[traceType]) {
         dataArrayPositionsByTraceType[traceType] = [];
       }
--- a/src/components/containers/TraceAccordion.js
+++ b/src/components/containers/TraceAccordion.js
@@ -10,78 +10,155 @@ import {Tab, Tabs, TabList, TabPanel} from 'react-tabs';
 const TraceFold = connectTraceToPlot(PlotlyFold);
 
 class TraceAccordion extends Component {
-  render() {
-    const {data = [], localize: _} = this.context;
-    const {
-      canAdd,
-      canGroup,
-      children,
-      messageIfEmptyFold,
-      excludeFits,
-    } = this.props;
+  renderGroupedTraceFolds(groupedTraces) {
+    if (this.props.useFullData) {
+      const dataArrayPositionsByTraceType = {};
+      const fullDataArrayPositionsByTraceType = {};
 
-    // we don't want to include analysis transforms when we're in the create panel
-    const filteredData = data.filter(t => {
-      if (excludeFits) {
-        return !(t.transforms && t.transforms.every(tr => tr.type === 'fit'));
-      }
-      return true;
-    });
+      Object.keys(groupedTraces).forEach(traceType => {
+        const dataIndexes = [];
+        const fullDataIndexes = [];
 
-    const individualTraces =
-      filteredData.length &&
-      filteredData.map((d, i) => {
+        this.context.fullData.forEach((trace, fullDataIndex) => {
+          if (trace.type === traceType && trace._expandedIndex) {
+            fullDataIndexes.push(trace._expandedIndex);
+          }
+          if (trace.type === traceType && !trace._expandedIndex) {
+            fullDataIndexes.push(fullDataIndex);
+          }
+          if (trace.type === traceType) {
+            dataIndexes.push(trace.index);
+          }
+        });
+
+        dataArrayPositionsByTraceType[traceType] = dataIndexes;
+        fullDataArrayPositionsByTraceType[traceType] = fullDataIndexes;
+      });
+
+      return Object.keys(groupedTraces).map((traceType, index) => {
         return (
           <TraceFold
-            key={i}
-            traceIndexes={[i]}
-            canDelete={canAdd}
-            messageIfEmpty={messageIfEmptyFold}
+            key={index}
+            traceIndexes={dataArrayPositionsByTraceType[traceType]}
+            name={traceType}
+            fullDataArrayPosition={fullDataArrayPositionsByTraceType[traceType]}
           >
-            {children}
+            {this.props.children}
           </TraceFold>
         );
       });
-
-    if (canAdd) {
-      const addAction = {
-        label: _('Trace'),
-        handler: ({onUpdate}) => {
-          if (onUpdate) {
-            onUpdate({
-              type: EDITOR_ACTIONS.ADD_TRACE,
-            });
-          }
-        },
-      };
-      return (
-        <PlotlyPanel addAction={addAction}>
-          {individualTraces ? individualTraces : null}
-        </PlotlyPanel>
-      );
     }
-    const tracesByGroup = filteredData.reduce((allTraces, nextTrace, index) => {
-      const traceType = plotlyTraceToCustomTrace(nextTrace);
-      if (!allTraces[traceType]) {
-        allTraces[traceType] = [];
-      }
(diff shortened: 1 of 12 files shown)


src/components/containers/TraceAccordion.js
```diff
@@ -10,78 +10,155 @@ import {Tab, Tabs, TabList, TabPanel} from 'react-tabs';
...
-    const {data = [], localize: _} = this.context;
-    const {
-      canAdd,
-      canGroup,
-      children,
-      messageIfEmptyFold,
-      excludeFits,
-    } = this.props;
+  renderGroupedTraceFolds(groupedTraces) {
+    if (this.props.useFullData) {
```
nicolaskruchten: Let's merge `useFullData` and `canGroup` so that you don't need this `if`... mostly so you can remove the `else` case below :)

src/components/containers/TraceAccordion.js
```diff
@@ -10,78 +10,155 @@ import {Tab, Tabs, TabList, TabPanel} from 'react-tabs';
...
-        label: _('Trace'),
-        handler: ({onUpdate}) => {
-          if (onUpdate) {
-            onUpdate({
-              type: EDITOR_ACTIONS.ADD_TRACE,
-            });
-          }
-        },
-      };
+    return Object.keys(groupedTraces).map((traceType, index) => {
```
nicolaskruchten: this is the "grouped but not full data" case that's never used

src/components/containers/TraceAccordion.js
```diff
@@ -10,78 +10,155 @@ import {Tab, Tabs, TabList, TabPanel} from 'react-tabs';
...
+          name={traceType}
+        >
+          {this.props.children}
+        </TraceFold>
       );
+    });
+  }
+
+  renderIndividualTraceFolds(filteredTraces, filteredTracesFullDataPositions) {
+    if (this.props.useFullData) {
```
nicolaskruchten: Good refactoring! It shows that there's basically no reuse in this function any more, so we should go the next step and split it in half: `renderTraceFolds` for the non-`canGroup` case and `renderUnGroupedTraceFolds` for the `canGroup`-but-ungrouped case.

src/components/containers/TraceAccordion.js
```diff
@@ -10,78 +10,155 @@ import {Tab, Tabs, TabList, TabPanel} from 'react-tabs';
...
+          allTraces[traceType] = [];
+        }
+        allTraces[traceType].push(adjustedIndex);
+        return allTraces;
+      },
+      {}
+    );
 
-    if (canGroup && filteredData.length > 1 && groupedTraces.length > 0) {
+    if (canAdd) {
```
nicolaskruchten: This is basically an early-return case that means we don't need `groupedTraces` so let's move it up above the computation of `groupedTraces` and save a bit of time when `canAdd`

src/components/containers/TraceAccordion.js
```diff
@@ -91,19 +168,34 @@ class TraceAccordion extends Component {
             </TabList>
             <TabPanel>
               <PlotlyPanel>
-                {individualTraces ? individualTraces : null}
+                {filteredTraces.length
```
nicolaskruchten: this ternary is reused in a few places. Let's push it into `renderIndividualTraceFolds` to make it a bit more robust, and then we can make this code more readable.

src/components/containers/TraceAccordion.js
```diff
@@ -91,19 +168,34 @@ class TraceAccordion extends Component {
...
+                      filteredTraces,
+                      filteredTracesFullDataPositions
+                    )
+                  : null}
               </PlotlyPanel>
             </TabPanel>
             <TabPanel>
-              <PlotlyPanel>{groupedTraces ? groupedTraces : null}</PlotlyPanel>
+              <PlotlyPanel>
+                {Object.keys(groupedTraces).length
```
nicolaskruchten: let's push this ternary into `renderGroupedTraceFolds`

src/components/fields/DataSelector.js
```diff
@@ -83,6 +83,40 @@ export class UnconnectedDataSelector extends Component {
           : null,
       }
     );
+
+    if (
```
nicolaskruchten: it feels like this should be in `shame.js` no? Must it really be in this file?

src/default_panels/GraphSubplotsPanel.js
```diff
@@ -17,7 +17,7 @@ import {
 import {TRACE_TO_AXIS} from '../lib/constants';
 
 const GraphSubplotsPanel = (props, {localize: _}) => (
-  <SubplotAccordion canGroup>
+  <SubplotAccordion>
```
nicolaskruchten: 👍

src/components/containers/TraceAccordion.js
```diff
@@ -10,38 +10,109 @@ import {Tab, Tabs, TabList, TabPanel} from 'react-tabs';
...
+          canDelete={this.props.canAdd}
+          messageIfEmpty={this.props.messageIfEmptyFold}
+        >
+          {this.props.children}
+        </TraceFold>
+      );
+    });
+  }
+
+  render() {
```
nicolaskruchten: Nice refactoring. This is a pretty hairy component and it's already looking a lot better! At some point I wonder if we can't break it apart more completely by having a `GroupedTraceAccordion` but not today.

nicolaskruchten:
Percy's diffs seem really odd to me here, not sure what it's doing. Locally I see Trace Opacity just fine, and I don't see weird extra padding.

nicolaskruchten:
Only issue I see so far is:

image

src/shame.js
```diff
@@ -112,3 +112,55 @@ export const shamefullyAddTableColumns = (graphDiv, {traceIndexes, update}) => {
...
+  traceIndex,
+  splitTraceGroup
+) => {
+  if (!Array.isArray(splitTraceGroup)) {
+    splitTraceGroup = [splitTraceGroup]; // eslint-disable-line
+  }
+
+  let indexOfSplitTransform = null;
+
+  graphDiv.data[traceIndex].transforms.filter((t, i) => {
```
dmt0: why filter?

src/shame.js
```diff
@@ -112,3 +112,55 @@ export const shamefullyAddTableColumns = (graphDiv, {traceIndexes, update}) => {
...
+    if (t.type === 'groupby') {
+      indexOfSplitTransform = i;
+    }
+  });
+
+  function getProp(group) {
+    let indexOfStyleObject = null;
+
+    graphDiv.data[traceIndex].transforms[indexOfSplitTransform].styles.filter(
```
dmt0: and here

src/shame.js
VeraZab: lol, just didn't want to write a for loop.. i could write a for loop

src/shame.js
VeraZab: same

src/components/containers/TransformAccordion.js
```diff
@@ -83,9 +83,17 @@ class TransformAccordion extends Component {
...
       ));
 
+    // cannot have 2 Split transforms on one trace:
+    // https://github.com/plotly/plotly.js/issues/1742
+    const addActionOptions =
+      container.transforms &&
+      container.transforms.some(t => t.type === 'groupby')
+        ? transformTypes.filter(t => t.type !== 'groupby')
+        : transformTypes;
+
```
dmt0: I don't think you need to 'some' it, just go straight to 'filter' - some takes performance too!

src/shame.js
dmt0: forEach?

dmt0: our linter won't complain, but generally function inside filter should return something...

VeraZab: yeah you're right, changing

dmt0:
just comments above, and 💃!

VeraZab:
ok, so, from a conversation with @alexcjohnson , this seems to be the correct behaviour.
I don't know why we had that widget before, but it does not seem to be correct.

```
scatter traces with `fill` are not allowed to have opacity. The reason is that lines & markers need to be drawn in a separate group from the fills, in order to get the layering right when there are multiple filled traces, so it’s not possible to change the opacity of the entire trace as a single group.
```

Issue #636, opened 2018-08-06
Hide SVG/WebGL behind a gear


nicolaskruchten:
Just a simple state: click once to show, again to hide.

Make it look like this: 

![image](https://user-images.githubusercontent.com/203523/43737515-8947b9d0-998f-11e8-8929-2b1516e8d6fd.png)

PR #639, opened 2018-08-06
hide gl controls behind a cogwheel icon

dmt0:
closes https://github.com/plotly/react-chart-editor/issues/636
giphy

### DIFF
--- a/src/components/fields/TraceSelector.js
+++ b/src/components/fields/TraceSelector.js
@@ -7,12 +7,13 @@ import {
   plotlyTraceToCustomTrace,
   computeTraceOptionsFromSchema,
 } from 'lib';
-import TraceTypeSelector, {
+import {
+  TraceTypeSelector,
   TraceTypeSelectorButton,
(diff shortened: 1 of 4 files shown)


src/components/fields/TraceSelector.js
```diff
@@ -120,18 +128,36 @@ class TraceSelector extends Component {
...
-              handleClick={() =>
-                this.context.openModal(TraceTypeSelector, {
-                  ...props,
-                  glByDefault: this.context.glByDefault,
-                })
-              }
-            />
+            <div
+              style={{
+                display: 'flex',
```
VeraZab: could we put this into a stylesheet :)

VeraZab:
yup, looks good 💃 
after small comment

src/components/fields/TraceSelector.js
dmt0: this style will not be reused anywhere, the main purpose of it is to override `field__widget`'s flex-wrap style

PR #642, opened 2018-08-07
color picker inline

### LINKED ISSUE #626: ColorPicker scroll behaviour
* If a ColorPicker is open and your mouse is *not* over the picker, you can't scroll the pane.
* If a ColorPicker opens at the bottom of a panel, it's not visible and it's hard to scroll into view

I think ideally the colorpicker should open inline like the new colorscale behaviour, instead of popping over.

/cc @cldougl

dmt0:
Closes #626

### DIFF
--- a/src/components/fields/MarkerColor.js
+++ b/src/components/fields/MarkerColor.js
@@ -110,23 +110,21 @@ class UnconnectedMarkerColor extends Component {
   renderConstantControls() {
     const _ = this.context.localize;
     return (
-      <div className="markercolor-constantcontrols__container">
-        <MultiColorPicker
-          attr="marker.color"
-          multiColorMessage={_(
-            'Each trace will be colored according to the selected colorscale.'
-          )}
-          singleColorMessage={_(
-            'All traces will be colored in the the same color.'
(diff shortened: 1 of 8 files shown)


VeraZab:
in Percy's `panelTest_StyleColorbarsPanel` test, colorPickers look misaligned

VeraZab:
there seems to be less space.. does this still look ok with an open colorscalepicker?

VeraZab:
A more minimalistic picker, nice, I like!
screen shot 2018-08-07 at 4 53 20 pm

nicolaskruchten:
@VeraZab yeah, this is because it's a fold in a fold and the color pickers are now fixed-width. I think we should accept this one for now :)

VeraZab:
ok well codewise looks ok to me
and could we open an issue for the misaligned colorpickers for another time?

otherwise 💃

dmt0:
Issue for misaligned colorpickers raised:
https://github.com/plotly/react-chart-editor/issues/648

nicolaskruchten:
💃 from me, including the Percy diffs

PR #645, opened 2018-08-07
DataSelector edge cases

### LINKED ISSUE #235: Behaviour for traces without *src
Right now if I load up an editor with a valid figure that has a trace with `x`  and `y` but no `xrsc` or `ysrc` I'll see something like this:

image

That seems... incomplete. Suggestions?

### LINKED ISSUE #602: Disable DataSelector if no data sources are present
When you load up a saved figure in https://github.com/plotly/jupyterlab-chart-editor there are no data sources per se: all the data is inlined. To properly support this use case, we need to disable the data-selectors.

nicolaskruchten:
* closes #602 
* closes #235

### DIFF
--- a/src/components/fields/DataSelector.js
+++ b/src/components/fields/DataSelector.js
@@ -48,6 +48,8 @@ export class UnconnectedDataSelector extends Component {
         (props.container.type === 'table' && props.attr !== 'columnorder') ||
         Array.isArray(this.fullValue);
     }
+
+    this.hasData = props.attr in props.container;
   }
 
   updatePlot(value) {
@@ -108,7 +110,10 @@ export class UnconnectedDataSelector extends Component {
           optionRenderer={this.context.dataSourceOptionRenderer}
           valueRenderer={this.context.dataSourceValueRenderer}
           clearable={true}
-          placeholder={this.props.placeholder}
+          placeholder={
+            this.hasData ? 'Data inlined in figure' : 'Choose data...'
+          }
+          disabled={this.dataSourceOptions.length === 0}
         />
       </Field>
     );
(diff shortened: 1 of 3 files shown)


src/components/fields/DataSelector.js
```diff
@@ -108,7 +110,10 @@ export class UnconnectedDataSelector extends Component {
           optionRenderer={this.context.dataSourceOptionRenderer}
           valueRenderer={this.context.dataSourceValueRenderer}
           clearable={true}
-          placeholder={this.props.placeholder}
+          placeholder={
+            this.hasData ? 'Data inlined in figure' : 'Choose data...'
```
VeraZab: well, the placeholder will never show right, it this.hasData is true..
because instead there will be something in the Data selector..

but ok

src/components/fields/DataSelector.js
```diff
@@ -108,7 +110,10 @@ export class UnconnectedDataSelector extends Component {
...
+          }
+          disabled={this.dataSourceOptions.length === 0}
```
VeraZab: nice!

src/components/fields/DataSelector.js
nicolaskruchten: this is for the case when you're opening a plot in JupyterLab without any data sources but inlined data.

src/components/fields/DataSelector.js
```diff
@@ -119,7 +124,6 @@ UnconnectedDataSelector.propTypes = {
   fullValue: PropTypes.any,
   updatePlot: PropTypes.func,
   container: PropTypes.object,
-  placeholder: PropTypes.string,
```
VeraZab: but don't we still want to have that capability of adding our custom string..
i guess we'll readjust when that becomes needed

nicolaskruchten: yep.

VeraZab:
ok well 💃

PR #650, opened 2018-08-08
move subplot-related widgets to subplot panel

### LINKED ISSUE #580: Subplots for all trace types

### LINKED ISSUE #624: Type-specific fields appearing inappopriately
See

image

dmt0:
closes #580 and closes #624


nicolaskruchten:
OK, with my latest push this commit does close #624 with respect to Radius. There's an old `r` field in `scatter` for some reason, and there was a bug in `TraceTypeSection` which wasn't protecting us correctly.

dmt0:
Looks awesome! 💃

Issue #653, opened 2018-08-08
scattergeo lat/lon OR locations


nicolaskruchten:
Right now when you create a `scattergeo` trace (aka Atlas Map) you see this:

![image](https://user-images.githubusercontent.com/203523/43844088-eb329f76-9af6-11e8-8aa3-517017cda395.png)

But in fact you can't use lat/lon and locations at the same time, it's one or the other. Furthermore when you give locations, the `locationmode` attr appears. This `locationmode` is also used in `choropleth` though.

So we need:

1. A super-widget that handles `locations` and `locationmode` together and shares visibility
2. A super-super-widget that allows users of `scattergeo` to say "I want to give my data in lat/lon" or "I want to give my data by location"

PR #655, opened 2018-08-08
location or lat/lon, not both

dmt0:
closes https://github.com/plotly/react-chart-editor/issues/653


VeraZab:
code looks good 💃

dmt0:
made a tiny change - not exporting all those components that are not useful outside the LocationSelector

PR #666, opened 2018-08-09
correct hoveron options for each trace type

dmt0:
closes https://github.com/plotly/react-chart-editor/issues/576

### DIFF
--- a/src/components/fields/derived.js
+++ b/src/components/fields/derived.js
@@ -539,16 +539,12 @@ export const HoverInfo = connectToContainer(UnconnectedFlaglist, {
       ].includes(container.type)
     ) {
       options.push({label: _('Z'), value: 'z'});
-    }
-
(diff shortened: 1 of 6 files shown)


src/components/fields/derived.js
```diff
@@ -539,16 +539,12 @@ export const HoverInfo = connectToContainer(UnconnectedFlaglist, {
       ].includes(container.type)
     ) {
       options.push({label: _('Z'), value: 'z'});
-    }
-
-    if (container.type === 'choropleth') {
+    } else if (container.type === 'choropleth') {
```
VeraZab: aww.. sorry i missed that in the review :S

VeraZab: did you change this to else if because the same trace type was present in more than one conditional?

VeraZab:
ok looks good to me 💃

PR #670, opened 2018-08-10
colorpicker css tweaks and version bump

### LINKED ISSUE #629: Categorical color scales should return full range
Right now the categorical color scales only return 7 colors even if they have 20. They should display and return their full range.

nicolaskruchten:
Closes #629

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "prop-types": "^15.5.10",
     "raf": "^3.4.0",
     "react-color": "^2.13.8",
-    "react-colorscales": "0.5.9",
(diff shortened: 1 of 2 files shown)


VeraZab:
I like it!
I'd say it looks better even with those variable colorswatch widths
screen shot 2018-08-10 at 9 36 27 am

Maybe the only little odd thing is cubhelix naming..
screen shot 2018-08-10 at 9 35 20 am

Could it maybe be like?:
s3 r1.5
s0 r0.4

But otherwise 💃 and thanks for aligning that widget :)

PR #675, opened 2018-08-12
Color cleanup

### LINKED ISSUE #529: Add Colorway picker
Break out Colorscale, Colorway, PieColors and the thing that's imported in MarkerColor ... much simpler!

nicolaskruchten:
Closes #529


src/components/fields/PieColorscalePicker.js
```diff
@@ -0,0 +1,80 @@
...
+      plotProps.fullValue = context.graphDiv.calcdata[0].map(d => [0, d.color]);
+    }
+
+    if (
+      plotProps.fullValue &&
+      Array.isArray(plotProps.fullValue) &&
+      !plotProps.fullValue.every(el => Array.isArray(el))
+    ) {
+      plotProps.fullValue = plotProps.fullValue.map(c => [0, c]);
+    }
```
VeraZab: same here, I think this whole conditional here is not needed for PieColorscalePickers

src/components/fields/PieColorscalePicker.js
```diff
@@ -0,0 +1,80 @@
...
+
+    if (context.traceIndexes.length > 1) {
+      plotProps.isVisible = false;
+    }
```
VeraZab: this is the conditional that didn't allow us to show the widget for multiple pie traces, is this still what we want?

src/components/fields/MultiColorPicker.js
```diff
@@ -8,6 +8,24 @@ import React, {Component} from 'react';
...
+      context.fullData
+        .filter(t => context.traceIndexes.includes(t.index))
+        .every(t => t.marker && t.marker.color) &&
+      (plotProps.fullValue && typeof plotProps.fullValue === 'string')
+    ) {
+      plotProps.fullValue =
+        context.fullData &&
+        context.fullData
+          .filter(t => context.traceIndexes.includes(t.index))
+          .map(t => [0, t.marker.color]);
```
VeraZab: so for the same reason as for pies below, I think we no longer want the `.map`

src/components/fields/ColorwayPicker.js
```diff
@@ -0,0 +1,31 @@
+import ColorscalePickerWidget from '../widgets/ColorscalePicker';
```
VeraZab: yup, that's a really nice one to expose :)

src/components/fields/PieColorscalePicker.js
nicolaskruchten: Yeah I think so

src/components/fields/PieColorscalePicker.js
```diff
@@ -0,0 +1,80 @@
...
+  modifyPlotProps: (props, context, plotProps) => {
+    if (
+      context &&
+      context.container &&
+      context.graphDiv &&
+      (!plotProps.fullValue ||
+        (Array.isArray(plotProps.fullValue) && !plotProps.fullValue.length)) &&
+      context.graphDiv.calcdata
+    ) {
+      plotProps.fullValue = context.graphDiv.calcdata[0].map(d => [0, d.color]);
```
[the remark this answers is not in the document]
nicolaskruchten: This used to have something like “&& is a pie” so I assumed it was still needed in this specific case...

src/components/fields/MultiColorPicker.js
nicolaskruchten: Again this used to be “&& is marker.color” so I think it’s still needed

src/components/fields/ColorwayPicker.js
nicolaskruchten: Yeah :)

What do you think of the various renamings? I did what I could to make them clearer... we used to import ColorscalePicker from Colorscale.js which has bugged me for months ;)

src/components/fields/PieColorscalePicker.js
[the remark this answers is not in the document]
VeraZab: it's needed :) just the `.map` is not needed, because we have the widget component that just takes an array of colors, and the attribute connected components now take their data in their respective formats.

src/components/fields/MultiColorPicker.js
VeraZab: same as above, my comment was just about the `.map` part :)

src/components/fields/ColorwayPicker.js
VeraZab: yup, the names are good and clear to me!
great job on separating them all :)

VeraZab:
💃

PR #676, opened 2018-08-12
Fix various visibility issues in Axes

### LINKED ISSUE #441: Remove TraceTypeSection from AxesFolds
We need a new AxesTypeSection for this purpose. This would be useful for fixing the problem we have with the `size` attr which means different things in the context of radial ("direction") vs cartesian ("position") axes

### LINKED ISSUE #510: axes tab: remove position field from axes/lines/fold

### LINKED ISSUE #511: axes tab: when hide option is selected in certain folds, then non relevant options should be hidden

### LINKED ISSUE #512: axes: move lines/mirror to layout

### LINKED ISSUE #535: side in StyleAxesPanel is overloaded in tick labels for cartesian (bottom/top/left/right) and polar (direction CW/CCW) so both appear and it's weird.

### LINKED ISSUE #672: Axis "side" controls
Right now in the Axes panel we have controls for various "Sides" that always show bottom/top/left/right and it's weird because in X mode it should be just left/right and in Y mode just bottom/top. In "grouped" mode that widget should just not appear as there's no option that makes sense unless we build a complex custom component that controls x and y separately which I don't want to do :)

### LINKED ISSUE #673: Mirror options
Right now we just have on/off but there are other options: https://plot.ly/javascript/reference/#layout-xaxis-mirror

nicolaskruchten:
Closes #535 
Closes #511 
Closes #510 
Closes #512 
Closes #441 
Closes #673 
Closes #672 

@dmt0 please review this guy as it basically implements a whole bunch of stuff you and I talked about and also changes the behaviour of `VisibilitySelect` a fair bit :)

### DIFF
--- a/src/components/fields/MarkerColor.js
+++ b/src/components/fields/MarkerColor.js
@@ -223,7 +223,7 @@ class UnconnectedMarkerColor extends Component {
                   {label: _('Custom'), value: false},
                 ]}
                 showOn={false}
-                dafault={true}
+                defaultOpt={true}
               >
                 <Numeric label={_('Min')} attr="marker.cmin" />
                 <Numeric label={_('Max')} attr="marker.cmax" />
--- a/src/components/fields/VisibilitySelect.js
+++ b/src/components/fields/VisibilitySelect.js
@@ -1,31 +1,39 @@
 import React, {Fragment, Component} from 'react';
 import PropTypes from 'prop-types';
 import {connectToContainer} from 'lib';
+import {MULTI_VALUED_PLACEHOLDER} from 'lib/constants';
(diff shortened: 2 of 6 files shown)


dmt0:
Looks good to me! 💃

nicolaskruchten:
Ah right I think I just decided to leave it in place and close the issue,
after playing with like 4 options ;)

dmt0:
How about those `true | "ticks" | false | "all" | "allticks"` - no use?

nicolaskruchten:
No, too complicated... I should switch from true/false to “allticks”/false
tho you’re right, I forgot

PR #678, opened 2018-08-13
dropdown with custom field

### LINKED ISSUE #513: axes label prefix/suffix: should have a input field to add custom prefix when on 'custom' option

### LINKED ISSUE #518: collapse-all missing in color bars folds

### LINKED ISSUE #533: extra padding appears when empty Panels are rendered in colorbar folds

### LINKED ISSUE #668: marker.colorbar.tickprefix custom prefix needs to be implemented
When the dropdown is set to custom, a text field should slide out.

dmt0:
closes #513 
closes #668 
closes #518 
closes #533


src/default_panels/StyleColorbarsPanel.js
```diff
@@ -185,49 +186,54 @@ const StyleColorBarsPanel = (props, {localize: _}) => {
...
+                  {label: _('$'), value: '$'},
+                  {label: _('#'), value: '#'},
+                  {label: _('@'), value: '@'},
+                  {label: _('custom'), value: 'custom'},
+                ]}
+                customOpt="custom"
+                dafaultOpt=""
+                clearable={false}
+              />
+              <Radio
```
nicolaskruchten: thanks for removing the sections... please add a label here like I did in style/axes

src/components/fields/DropdownCustom.js
```diff
@@ -0,0 +1,95 @@
...
+          value={value}
+          onChange={this.setValue}
+          clearable={this.props.clearable}
+          optionRenderer={this.props.optionRenderer}
+          valueRenderer={this.props.valueRenderer}
+          placeholder={this.props.placeholder}
+        />
+
+        {this.state.custom && (
+          <Text attr={attr} updatePlot={value => this.setValue(value, true)} />
```
nicolaskruchten: this doesn't seem to update as I type, only when I hit 'enter'

nicolaskruchten: Also, we could just have this send back `"custom"` when the field is empty, and preload it with `"custom"` so that the next field doesn't pop in and out of existence.

src/default_panels/StyleColorbarsPanel.js
dmt0: done

nicolaskruchten:
UX issue: if I choose a custom prefix like "lalala" and then close/reopen the fold, the state is blown away and it shows up as "please choose". This type of thing is why we basically can never use `state` in our custom components... it's too brittle. We should do something like iterate over the options and assume that if the prefix is not one of the options, it's a custom one (with a special case for "" with state if needed)

dmt0:
addressed all the above

VeraZab:
looks good to me 💃

PR #680, opened 2018-08-13
moving attr

nicolaskruchten:


dmt0:
once circleCI thing is fixed - 💃 !

PR #682, opened 2018-08-14
fix colorpicker within fold within fold style

### LINKED ISSUE #648: Colorpickers are misaligned when inside fold inside a fold
Such as on Colorbars panel

dmt0:
closes #648


nicolaskruchten:
💃

Issue #631, opened 2018-08-06
Refactor: merge AxesCreator and SubplotCreator


dmt0:
Most of the functionality is same, there's a lot of code duplication between the two.
Also getAllAxes and getAllSubplots.

PR #687, opened 2018-08-14
localize subplot names

### LINKED ISSUE #665: Localized subplot/axis names in subplot panel folds and elsewhere

dmt0:
closes #665 
addresses #631


nicolaskruchten:
Looking good, nice refactoring/localization!

Small things:

1. Pie seems to get its number from the trace number in general, so if the first pie is the 7th trace, the subplot is Pie 6
2. The small icons for cones/streamtubes are still defaulting to the 'line' icon... can we patch things so they get the corresponding temp icons for scatter3d and scatterline please?

image

src/components/containers/SubplotAccordion.js
```diff
@@ -111,23 +118,34 @@ class SubplotAccordion extends Component {
...
-            name={layoutKey}
+            name={subplotName}
           >
             {children}
           </NonCartesianSubplotFold>
         );
       }
     });
 
+    let pieCounter = 0;
```
nicolaskruchten: hehe OK, we'll refactor that next time, when we add a third :)

nicolaskruchten:
:dancer:

src/components/containers/SubplotAccordion.js
dmt0: that's what I was thinking when I was writing it lol

Issue #562, opened 2018-07-06
When the last trace is deleted, annotation, shapes, images, etc. are still in figure.layout, but cannot be modified through UI


dmt0:
They should either get deleted together with last trace or should remain accessible.

nicolaskruchten:
A radical thought: what happens if we get rid of TraceRequired (i.e. just use normal panels) for panels other than Style/Traces?

PR #692, opened 2018-08-14
Empty panels

### LINKED ISSUE #688: TraceRequiredPanel should have link to create
"Create" should be clickable and cause navigation to the create panel. This should be a generic-ish solution as we'll want to use it in a couple of other places

image

### LINKED ISSUE #689: Sliders/Menus panel shouldn't show when none present
They should be hidden from the sidebar menu if the corresponding array in the figure is empty. I'm thinking some sort of attr on the panel that the panelmenu can check.

Might be convenient to do this in parallel with #688 as they'll both likely touch the panelmenu code :)

dmt0:
closes #688 
closes #689

### DIFF
--- a/src/DefaultEditor.js
+++ b/src/DefaultEditor.js
@@ -17,8 +17,58 @@ import {
   StyleUpdateMenusPanel,
 } from './default_panels';
 import Logo from './components/widgets/Logo';
+import {TRANSFORMABLE_TRACES} from './lib/constants';
 
 class DefaultEditor extends Component {
+  constructor(props, context) {
+    super(props, context);
+    this.hasTransforms = this.hasTransforms.bind(this);
+    this.hasAxes = this.hasAxes.bind(this);
+    this.hasMenus = this.hasMenus.bind(this);
+    this.hasSliders = this.hasSliders.bind(this);
+    this.hasColorbars = this.hasColorbars.bind(this);
+  }
+
+  hasTransforms() {
+    return this.context.fullData.some(d =>
+      TRANSFORMABLE_TRACES.includes(d.type)
+    );
+  }
+
+  hasAxes() {
+    return (
+      Object.keys(this.context.fullLayout._subplots).filter(
+        type =>
+          !['cartesian', 'mapbox'].includes(type) &&
+          this.context.fullLayout._subplots[type].length > 0
+      ).length > 0
+    );
+  }
+
+  hasMenus() {
+    const {
+      fullLayout: {updatemenus = []},
+    } = this.context;
+
+    return updatemenus.length > 0;
+  }
+
+  hasSliders() {
+    const {
+      layout: {sliders = []},
+    } = this.context;
+
+    return sliders.length > 0;
+  }
+
+  hasColorbars() {
+    return this.context.fullData.some(
+      d =>
+        (d.marker && d.marker.showscale !== undefined) || // eslint-disable-line no-undefined
+        d.showscale !== undefined // eslint-disable-line no-undefined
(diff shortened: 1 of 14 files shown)


src/components/containers/TraceRequiredPanel.js
```diff
@@ -12,55 +12,36 @@ class TraceRequiredPanel extends Component {
...
     const {children, ...rest} = this.props;
     let showPanel = true;
-    const emptyPanelMessage = {heading: '', message: ''};
+    let heading = '';
 
-    const noTraceMessage = {
-      heading: _("Looks like there aren't any traces defined yet."),
-      message: _("Go to the 'Create' tab to define traces."),
-    };
+    if (!this.props.visible || !showPanel) {
```
nicolaskruchten: Nice refactoring! Please do one more pass here though... you can get rid of `showPanel` I think, and inline the "Looks like there aren't any traces defined yet" heading and so on :)

src/components/containers/TraceRequiredPanel.js
```diff
@@ -12,55 +12,36 @@ class TraceRequiredPanel extends Component {
...
-      );
-    }
-    return null;
   }
 }
 
 TraceRequiredPanel.propTypes = {
   children: PropTypes.node,
   visible: PropTypes.bool,
-  extraConditions: PropTypes.array,
```
nicolaskruchten: ah nice.

src/components/containers/UpdateMenuAccordion.js
```diff
@@ -38,19 +38,7 @@ class UpdateMenuAccordion extends Component {
...
-          {
-            heading: _('There are no update menus to style in your plot'),
-            message: '',
-          },
-        ]}
-      >
-        {content ? content : null}
-      </TraceRequiredPanel>
-    );
+    return <TraceRequiredPanel>{content ? content : null}</TraceRequiredPanel>;
```
nicolaskruchten: Hmm ok once we tackle #562 this can just be a Panel :)

src/default_panels/StyleAxesPanel.js
```diff
@@ -22,33 +22,10 @@ import {
 } from '../components';
 
 class StyleAxesPanel extends Component {
-  constructor(props, context) {
```
nicolaskruchten: very nice

src/DefaultEditor.js
```diff
@@ -17,8 +17,58 @@ import {
...
+    } = this.context;
+
+    return sliders.length > 0;
+  }
+
+  hasColorbars() {
+    return this.context.fullData.some(
+      d =>
+        (d.marker && d.marker.showscale !== undefined) || // eslint-disable-line no-undefined
+        d.showscale !== undefined // eslint-disable-line no-undefined
```
nicolaskruchten: This is good. I know the panel still appears even when all traces are in constant-color mode, but that's ok. at least this panel never appears when there's nothing you can do in it :)

dmt0: once you saw that panel appear, you don't wanna loose it again right? :)

nicolaskruchten:
I'm loving this `setPanel` thing. we can use it judiciously in a couple of places to help the user navigate, like "go to the Color Bar panel to style it" or "go to the legend panel" or whatever

src/components/containers/TraceRequiredPanel.js
dmt0: yeah, that was a bit uncooked

nicolaskruchten:
💃 nice one!

PR #694, opened 2018-08-14
Panel tweaks

nicolaskruchten:
A new pass on Sliders and Menus, adding some basic positioning widgets and fixing a crash when hiding a menu.

### DIFF
--- a/src/default_panels/StyleSlidersPanel.js
+++ b/src/default_panels/StyleSlidersPanel.js
@@ -6,6 +6,7 @@ import {
   Numeric,
   Radio,
   PlotlySection,
+  Dropdown,
   SliderAccordion,
 } from '../components';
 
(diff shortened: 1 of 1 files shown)


src/default_panels/StyleSlidersPanel.js
```diff
@@ -31,6 +32,41 @@ const StyleSlidersPanel = (props, {localize: _}) => (
...
+      <Numeric label={_('Position')} attr={'x'} showSlider step={0.02} />
+      <Dropdown
+        label={_('Anchor')}
+        attr={'xanchor'}
+        options={[
+          {label: _('Left'), value: 'left'},
+          {label: _('Center'), value: 'center'},
+          {label: _('Right'), value: 'right'},
+        ]}
+      />
```
dmt0: I think it needs to be `clearable={false}`, same for other 3 dropdowns.

dmt0: other than that 💃

nicolaskruchten:
Ok, could you plz make that change and merge? That way it gets into a daily
build sooner :)

Issue #656, opened 2018-08-08
Transforms for all traces?


dmt0:

nicolaskruchten:
`Contour` and `heatmap` always have 2d input so they're out, but `histogram2d` actually doesn't accept 2d input, despite the name.

PR #699, opened 2018-08-16
Empty folds

### LINKED ISSUE #695: Don't show useless ColorBar and Transform folds
Now that we don't let users see the Color Bar or Transform panels at all if they're 100% useless we can go one final step further and just not render the folds for the traces where nothing is possible, so users should never ever see "no color bar for this trace" or "no transforms possible here" or whatever.

dmt0:
addresses #656 
closes #695 

also bumps plotly.js to 1.40.0

### DIFF
--- a/src/components/containers/TraceAccordion.js
+++ b/src/components/containers/TraceAccordion.js
@@ -23,19 +23,21 @@ class TraceAccordion extends Component {
   setLocals(props, context) {
     // we don't want to include analysis transforms when we're in the create panel
     const base = props.canGroup ? context.fullData : context.data;
+    const traceFilterCondition =
+      this.props.traceFilterCondition || (() => true);
 
-    this.filteredTracesFullDataPositions = [];
+    this.filteredTracesIndexes = [];
     this.filteredTraces = base.filter((t, i) => {
-      if (props.excludeFits) {
-        return !(t.transforms && t.transforms.every(tr => tr.type === 'fit'));
+      if (traceFilterCondition(t, context.fullData[i])) {
+        this.filteredTracesIndexes.push(i);
+        return true;
       }
-      this.filteredTracesFullDataPositions.push(i);
-      return true;
+      return false;
     });
   }
 
   renderGroupedTraceFolds() {
-    if (!this.filteredTraces.length || this.filteredTraces.length < 2) {
+    if (!this.filteredTraces.length || this.filteredTraces.length <= 1) {
       return null;
     }
 
@@ -55,61 +57,45 @@ class TraceAccordion extends Component {
 
       dataArrayPositionsByTraceType[traceType].push(trace.index);
       fullDataArrayPositionsByTraceType[traceType].push(
-        this.filteredTracesFullDataPositions[index]
+        this.filteredTracesIndexes[index]
       );
     });
 
-    return Object.keys(fullDataArrayPositionsByTraceType).map((type, index) => {
-      return (
-        <TraceFold
-          key={index}
-          traceIndexes={dataArrayPositionsByTraceType[type]}
-          name={traceTypes(_).find(t => t.value === type).label}
-          fullDataArrayPosition={fullDataArrayPositionsByTraceType[type]}
-        >
-          {this.props.children}
-        </TraceFold>
-      );
-    });
+    return Object.keys(fullDataArrayPositionsByTraceType).map((type, index) => (
+      <TraceFold
+        key={index}
+        traceIndexes={dataArrayPositionsByTraceType[type]}
+        name={traceTypes(_).find(t => t.value === type).label}
+        fullDataArrayPosition={fullDataArrayPositionsByTraceType[type]}
+      >
+        {this.props.children}
+      </TraceFold>
+    ));
   }
 
   renderUngroupedTraceFolds() {
-    if (!this.filteredTraces.length) {
-      return null;
-    }
-
-    return this.filteredTraces.map((d, i) => {
-      return (
-        <TraceFold
-          key={i}
-          traceIndexes={[d.index]}
-          canDelete={this.props.canAdd}
-          messageIfEmpty={this.props.messageIfEmptyFold}
-          fullDataArrayPosition={[this.filteredTracesFullDataPositions[i]]}
-        >
-          {this.props.children}
-        </TraceFold>
-      );
-    });
+    return this.filteredTraces.map((d, i) => (
+      <TraceFold
+        key={i}
+        traceIndexes={[d.index]}
+        canDelete={this.props.canAdd}
+        fullDataArrayPosition={[this.filteredTracesIndexes[i]]}
+      >
+        {this.props.children}
+      </TraceFold>
+    ));
   }
 
   renderTraceFolds() {
-    if (!this.filteredTraces.length) {
-      return null;
-    }
-
-    return this.filteredTraces.map((d, i) => {
-      return (
-        <TraceFold
-          key={i}
-          traceIndexes={[i]}
-          canDelete={this.props.canAdd}
-          messageIfEmpty={this.props.messageIfEmptyFold}
-        >
-          {this.props.children}
-        </TraceFold>
-      );
-    });
+    return this.filteredTraces.map((d, i) => (
+      <TraceFold
+        key={i}
+        traceIndexes={[this.filteredTracesIndexes[i]]}
+        canDelete={this.props.canAdd}
+      >
+        {this.props.children}
+      </TraceFold>
+    ));
   }
 
   render() {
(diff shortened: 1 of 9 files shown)


src/components/containers/TraceAccordion.js
```diff
@@ -59,63 +81,72 @@ class TraceAccordion extends Component {
...
 
   renderUngroupedTraceFolds() {
-    if (!this.filteredTraces.length) {
-      return null;
-    }
-
-    return this.filteredTraces.map((d, i) => {
-      return (
+    return (
+      this.filteredTraces.length > 0 &&
```
nicolaskruchten: I think this changes the meaning... now when `length === 0` we return `false` instead of `null`.

src/components/containers/TraceAccordion.js
```diff
@@ -24,14 +24,36 @@ class TraceAccordion extends Component {
...
+      ) {
+        return i;
+      } else if (
+        props.excludeNonTransformable &&
+        TRANSFORMABLE_TRACES.includes(t.type)
+      ) {
+        return i;
+      } else if (
+        props.excludeNoColorbar &&
+        ((context.fullData[i].marker &&
```
nicolaskruchten: this code is repeated in the PanelMenu piece that hides the ColorBar panel... can you think of a way to centralize it?

src/components/containers/TraceAccordion.js
```diff
@@ -24,14 +24,36 @@ class TraceAccordion extends Component {
     // we don't want to include analysis transforms when we're in the create panel
     const base = props.canGroup ? context.fullData : context.data;
 
-    this.filteredTracesFullDataPositions = [];
-    this.filteredTraces = base.filter((t, i) => {
-      if (props.excludeFits) {
-        return !(t.transforms && t.transforms.every(tr => tr.type === 'fit'));
+    this.filteredTracesFullDataPositions = base.map((t, i) => {
+      if (
+        props.excludeFits &&
```
nicolaskruchten: OK we have three filters here... can we just pass in one filter function and have the filter code defined at the call point rather than have this big if?

src/lib/constants.js
```diff
@@ -107,12 +107,14 @@ export const TRANSFORMS_LIST = ['filter', 'groupby', 'aggregate', 'sort'];
...
 export const TRANSFORMABLE_TRACES = [
   'scatter',
-  'bar',
   'scattergl',
-  'histogram',
-  'histogram2d',
   'box',
   'violin',
+  'bar',
+  'ohlc',
```
nicolaskruchten: so here we've just added ohlc and candlestick, right?

nicolaskruchten:
With a `pie` as the first trace and a `scatter` as the second I get this:

image

With this error in the console: 

```
Warning: Failed prop type: Invalid prop `traceIndexes[0]` of type `boolean` supplied to `TraceConnectedPlotlyFold`, expected `number`.
    in TraceConnectedPlotlyFold (created by TraceAccordion)
    in TraceAccordion (created by GraphTransformsPanel)
    in GraphTransformsPanel (created by DefaultEditor)
```

nicolaskruchten:
Also when I have `pie` then `scatter ` with a colorscale in the `scatter`, the Color Bar panel is present but empty.

src/components/containers/TraceAccordion.js
```diff
@@ -24,14 +24,36 @@ class TraceAccordion extends Component {
     // we don't want to include analysis transforms when we're in the create panel
     const base = props.canGroup ? context.fullData : context.data;
 
-    this.filteredTracesFullDataPositions = [];
-    this.filteredTraces = base.filter((t, i) => {
-      if (props.excludeFits) {
-        return !(t.transforms && t.transforms.every(tr => tr.type === 'fit'));
+    this.filteredTracesFullDataPositions = base.map((t, i) => {
+      if (
```
VeraZab: I don't agree with this logic.
When base is data, like in the Create panel, you cannot assume that the index of an item in the data array is the same as its index in the fullData array.

The case I can think of is in the Create Panel, which has base data,
if you add 1 split trace, and 1 non split case.

Data will be like: [{type: 'scatter'}, {type: 'pie'}]
FullData will be like: [{type: 'scatter', transforms: []}, {type: 'scatter', transforms: []}, {type: 'pie'} ]

In the Create panel, you'd be mapping on data, so you'd have a this.filteredTracesFulDataPositions of [0, 1], where you should have [0, 2].

nicolaskruchten:
Well, no every exclusion has to apply to every case of grouped vs un
grouped... this is why I would move the filter out to the call point

src/lib/constants.js
dmt0: yes, that's all that can be done safely, box and violin are already there

src/components/containers/TraceAccordion.js
dmt0: Yeah, but AFAIK behaviour stays the same

src/components/containers/TraceAccordion.js
```diff
@@ -23,14 +23,16 @@ class TraceAccordion extends Component {
...
     // we don't want to include analysis transforms when we're in the create panel
     const base = props.canGroup ? context.fullData : context.data;
+    const traceFilterCondition =
+      this.props.traceFilterCondition || (() => true);
 
     this.filteredTracesFullDataPositions = [];
     this.filteredTraces = base.filter((t, i) => {
-      if (props.excludeFits) {
-        return !(t.transforms && t.transforms.every(tr => tr.type === 'fit'));
+      if (traceFilterCondition(t, context.fullData[i])) {
```
VeraZab: context.fullData[i]
...well, base could be data or fullData, given this, do we still want to send context.fullData[i]?
or maybe just base[i] is ok?

src/components/containers/TraceAccordion.js
VeraZab: could we return null here if it doesn't meet this condition.

src/components/containers/TraceAccordion.js
```diff
@@ -59,57 +61,47 @@ class TraceAccordion extends Component {
...
+          traceIndexes={[this.filteredTracesFullDataPositions[i]]}
           canDelete={this.props.canAdd}
-          messageIfEmpty={this.props.messageIfEmptyFold}
         >
           {this.props.children}
         </TraceFold>
-      );
-    });
+      ))
+    );
```
VeraZab: same return null when doesn't meet condition

src/components/containers/TraceAccordion.js
```diff
@@ -59,57 +61,47 @@ class TraceAccordion extends Component {
...
-
-    return this.filteredTraces.map((d, i) => {
-      return (
+    return (
+      this.filteredTraces.length > 0 &&
+      this.filteredTraces.map((d, i) => (
         <TraceFold
           key={i}
-          traceIndexes={[i]}
+          traceIndexes={[this.filteredTracesFullDataPositions[i]]}
```
VeraZab: yeah... im not sure about this, here i think we'd want data indexes not fullData, why do we need that change?

traceIndexes should stay data related, that's what its meaning was throughout the code.

src/default_panels/StyleColorbarsPanel.js
```diff
@@ -16,11 +16,13 @@ import {
...
+export const traceHasColorbar = (trace, fullTrace) =>
+  (fullTrace.marker && fullTrace.marker.showscale !== undefined) || // eslint-disable-line no-undefined
+  fullTrace.showscale !== undefined; // eslint-disable-line no-undefined
+
 const StyleColorBarsPanel = (props, {localize: _}) => {
   return (
-    <TraceAccordion
-      messageIfEmptyFold={_('Need a color scale for a colorbar!')}
-    >
+    <TraceAccordion traceFilterCondition={traceHasColorbar}>
```
VeraZab: yeah, this is much better on the api

VeraZab:
Could you please look at the 2 comments with 😕 , those are my 2 potentially blocking comments.

src/components/containers/TraceAccordion.js
dmt0: we ARE sending base[i]. The second argument is a special case for Colorbar stuff. Doesn't get looked at by any other filter condition funcs.

src/components/containers/TraceAccordion.js
dmt0: i is not index of data, it's index of a filteredTraces array. Perhaps I should make a `filteredTracesDataPositions`?

src/components/containers/TraceAccordion.js
dmt0: what difference would that make?

src/components/containers/TraceAccordion.js
dmt0: actually, `filteredTracesFullDataPositions` are not relative to fullData, they're relative to base which could be anything

VeraZab:
💃

src/components/containers/TraceAccordion.js
nicolaskruchten: If `null` and `false` are both ignored by React then I'm fine with this. If React also ignores `[]` then we can just return the result of `.map()` without the guard.

nicolaskruchten:
💃 on behaviour.

PR #705, opened 2018-08-21
Split trace fix

### LINKED ISSUE #696: Variable size and text for split traces
The following works OK:

1. Create a scatter trace bill vs tips
2. Size it by size (note: this writes to data)
3. Set text=smoker
3. Split it by sex

The same thing in a different order doesn't work OK:

1. Create a scatter trace bill vs tips
2. Split it by sex
3. can't size it by size any more (not sure where it's writing but it's not the right place!)
4. can't set text any more (not sure where it's writing but it's not the right place!)

Maybe we should just prevent this, or maybe we should pick it up and write through to data or something... ? Not even sure what happens if you do it in grouped context.

/cc @VeraZab you're busy in the integration branch so let's get @dmt0 to look at this one and familiarize himself with the split trace issues, but just wanted to make sure you saw this

### LINKED ISSUE #702: Grid-align trace selector buttons
We need a solution for this:

image

Let's just pad things out as if every trace type had a two-line name for now.

dmt0:
Closes #696
Closes #702

- Hide the SymbolSelector when symbol is picked (https://github.com/plotly/streambed/issues/11224#issuecomment-413973225)

### DIFF
--- a/src/components/widgets/SymbolSelector.js
+++ b/src/components/widgets/SymbolSelector.js
@@ -81,7 +81,10 @@ export default class SymbolSelector extends Component {
         <div
           className="symbol-selector__item"
           key={value}
-          onClick={() => this.props.onChange(value)}
+          onClick={() => {
+            this.props.onChange(value);
+            this.togglePanel();
+          }}
         >
           <svg
             width="28"
--- a/src/lib/connectTraceToPlot.js
+++ b/src/lib/connectTraceToPlot.js
@@ -125,6 +125,9 @@ export default function connectTraceToPlot(WrappedComponent) {
             )
           : null;
 
+        const containsAnSrc =
+          Object.keys(update).filter(a => a.endsWith('src')).length > 0;
+
         if (Array.isArray(update)) {
           update.forEach((u, i) => {
             this.context.onUpdate({
@@ -136,7 +139,7 @@ export default function connectTraceToPlot(WrappedComponent) {
               },
             });
           });
-        } else if (splitTraceGroup) {
+        } else if (splitTraceGroup && !containsAnSrc) {
           this.props.traceIndexes.forEach((t, i) => {
             this.context.onUpdate({
               type: EDITOR_ACTIONS.UPDATE_TRACES,
--- a/src/styles/components/containers/_modalbox.scss
+++ b/src/styles/components/containers/_modalbox.scss
@@ -30,3 +30,9 @@
     position: relative;
   }
 }
+
+.field .modalbox {
+  width: 100%;
+  left: -1px;
+  top: 100%;
+}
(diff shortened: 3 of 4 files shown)


src/components/widgets/SymbolSelector.js
```diff
@@ -81,7 +81,10 @@ export default class SymbolSelector extends Component {
         <div
           className="symbol-selector__item"
           key={value}
-          onClick={() => this.props.onChange(value)}
+          onClick={() => {
+            this.props.onChange(value);
+            this.togglePanel();
```
VeraZab: ok, although eventually, for consistency we'll probably need to do the same as with the color picker, where it appeared in line.

VeraZab:
besides comment above, looks good!

src/components/widgets/SymbolSelector.js
dmt0: Yep, raised an issue for that! #706

VeraZab: thx

src/lib/connectTraceToPlot.js
```diff
@@ -136,7 +139,7 @@ export default function connectTraceToPlot(WrappedComponent) {
               },
             });
           });
-        } else if (splitTraceGroup) {
+        } else if (splitTraceGroup && !containsAnSrc) {
```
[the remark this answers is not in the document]
dmt0: So the array that goes into `marker.size` also gets split in that case, so in `fullData` you see two different arrays, so it looks like Multiple Values. I added a check to prevent it, but ONLY in case when we have only ONE TRACE split into a few. That seems like the safest solution to me at the moment.

The problem with this is that, say if you have `trace 0` sized by `ints` and split into two, and `trace 1` also sized by `ints` and not split, you'll have Multiple Values show up even though they're sized with the same array. If I prevent it from comparing arrays in this case, than different arrays will look like the same value.

A more complete solution would be to pick out arrays from `_expandedInput` into the `fullData` traces just for the sake of doing the comparison in the `connectTraceToPlot.js`. It'll have to be done recursively too. I'll make a separate issue for that.

@nicolaskruchten FYI

src/lib/connectTraceToPlot.js
```diff
@@ -90,7 +90,8 @@ export default function connectTraceToPlot(WrappedComponent) {
         fullData.forEach(t =>
           Object.keys(t).forEach(key =>
             setMultiValuedContainer(multiValuedFullContainer, t, key, {
-              searchArrays: true,
+              // don't search arrays if all traces are from splits
+              searchArrays: data.length > 1,
```
VeraZab: no.. I don't think I agree with this..
if data.length === 1, that doesn't mean that all traces are from splits..
I think this pr is ok without this adjustment, and we can think about this one some more in another pr.

PR #715, opened 2018-08-28
support for examples and tutorials

dmt0:
`PlotlyEditor` can receive a prop `chartHelp` of type `object` that looks like
```
{
  area: {helpDoc: 'https://someurl/', examplePlot: () => {}, feedQuery: 'area'},
  bar: {helpDoc: 'https://someurl/', examplePlot: () => {}, feedQuery: 'bar'},
  ...
}
```
If the prop exists, examples, tutorials, and links to community examples show up in Trace Selector.

Example is a callback that is meant to update the figure.

Also had to downgrade packages, because `npm run make:lib:cs` was producing errors - there were some unparced SASS variables. `cssnano` particularly changed a lot going into v4.


nicolaskruchten:
Can you add some examples of this into the dev app plz?

src/components/widgets/TraceTypeSelector.js
```diff
@@ -7,19 +7,22 @@ import {traceTypeToPlotlyInitFigure, renderTraceIcon, plotlyTraceToCustomTrace}
...
-          role="tooltip"
-          href={action.href}
-          target="_blank"
-        >
-          {action.icon}
-        </a>
-      ))
+    ? actionItems(item).map(
+        (action, i) =>
+          !action.onClick ? null : (
```
nicolaskruchten: this could just be `action.onClick && x` instead of `!action.onClick ? null : x`. Not a big deal but it jumped out at me for some reason

PR #729, opened 2018-09-02
use new cone/streamtube icons

nicolaskruchten:


dmt0:
are the fold icons showing right? if yes than 💃 !

PR #747, opened 2018-09-25
Fit traces adjustments

VeraZab:
closes: https://github.com/plotly/streambed/issues/11524
A few issues were discovered by this bug:
- https://github.com/plotly/react-chart-editor/pull/747/commits/acd90b76c99a499df9e12bead47ee28658b3c513: when trying to style 2 scatters, in grouped mode, they're marked as 'area' traces.
- https://github.com/plotly/react-chart-editor/pull/747/commits/a8a562450d345287fd71d18b14e315c3619a288c: the main cause of the streambed bug, [setMultivaluedContainer was mutating gd.data/gd._fullData](https://github.com/plotly/react-chart-editor/blob/master/src/lib/multiValues.js#L41). In the case of a figure like:
```
data = [scatterTrace, scatterTrace, fitOfScatter1, fitOfScatter2]
```
In Style/Traces/By Type, we'd have 1 `Area` TraceFold, and setMultivaluedContainer would pick the first trace: a scatter, and compare all other traces against it to see if they have a different value. The problem would occur when it would arrive at the `transforms` key, a normal scatter wouldn't have that key, and so we'd enter into a [gd mutating state](https://github.com/plotly/react-chart-editor/blob/master/src/lib/multiValues.js#L41), and start writing `MIXED_VALUES` into gd. To prevent this, simply using deepCopy.

- https://github.com/plotly/react-chart-editor/pull/747/commits/fc8f4b32753322209df834a1727761b4b2761445: cleans up index logic in TraceAccordion. The logic here got a bit murky once we've added the ability to filter certain trace types from accordions and styling of transformed traces. `this.filteredTracesDataIndexes` is always meant to indicate the position in the data array of `this.filteredTraces`, no matter their base (fullData or Data). We're using `this.filteredTracesDataIndexes` in the `traceIndexes` prop on TraceFolds, and those are going to be the indexes used on gd.data to make changes to our traces. `_expandedIndex` is what we can use on a `fullTrace` to get its position in `gd._fullData`.

-  https://github.com/plotly/react-chart-editor/pull/747/commits/a44d192a10a543b86bfbde8e3dbca54f60a06581: clear up logic in connectTraceToPlot. Reading in `_fullInput` is no longer necessary. This was added at the time that financial traces were transforms and as such, [did not have all of the required info in their `gd._fullData` to make all the styling widgets appear](https://github.com/plotly/react-chart-editor/commit/de94891f28fa9ae3321fa9968346384009984d2f#diff-f70d21591ac281a5774a1d4c0788bc12). This is no longer the case, and I think it's better to remove to simplify the logic.


VeraZab:
While I figure out what's wrong with the test @dmt0 @nicolaskruchten , this is ready for review and questions.

nicolaskruchten:
Can you drop the test plz to get the checkbox back? :)

src/lib/connectNonCartesianSubplotToLayout.js
```diff
@@ -23,32 +23,17 @@ export default function connectNonCartesianSubplotToLayout(WrappedComponent) {
...
-          */
-          if (trace.transforms && trace.transforms.every(t => t.type === 'fit')) {
-            fullData[i]._fullInput = fullData[i];
-          }
-
-          fullTrace = fullData[i]._fullInput;
-
-          break;
-        }
+      let fullTrace = fullData.filter(t => t && traceIndexes[0] === t.index)[0] || {};
```
nicolaskruchten: since this logic is in three places now, could we extract it out into a util or something?

src/components/containers/TraceAccordion.js
```diff
@@ -71,28 +81,54 @@ class TraceAccordion extends Component {
...
+        <TraceFold
+          key={i}
+          traceIndexes={[this.filteredTracesDataIndexes[i]]}
+          canDelete={this.props.canAdd}
+        >
+          {this.props.children}
+        </TraceFold>
+      ));
+    }
+    return null;
```
nicolaskruchten: why did you change this to `null` from `[]`?

nicolaskruchten: just to avoid the `.length` check below?

VeraZab: yeah... that was just when I was trying things to get the breaking test to work..
and I thought, well, since this function is called `renderTraceFolds` then it felt more reacty to return null if those shouldn't display..

nicolaskruchten:
💃

src/components/containers/TraceAccordion.js
nicolaskruchten: hehe I don't think that feels more reacty at all :) but ok.

PR #772, opened 2018-10-19
preset sizeref reasonably

### LINKED ISSUE #771: Compute initial sizeref intelligently based on range of data
Right now if you set scatter marker size to a column with numbers in the millions, you end up with giant markers until you scale them down. Chrome doesn't like that at all and hangs unfortunately, so we need to precompute sizeref.

nicolaskruchten:
Closes #771

### DIFF
--- a/src/shame.js
+++ b/src/shame.js
@@ -202,3 +202,13 @@ export const shamefullyDeleteRelatedAnalysisTransforms = (graphDiv, payload) => 
     }
   }
 };
+
+export const shamefullyAdjustSizeref = (gd, {update}) => {
+  const {'marker.size': size = null, 'marker.sizesrc': src = null} = update;
+  if (size && src) {
+    const DEFAULT_MAX_AREA_PX = 45;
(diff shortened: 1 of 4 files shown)


src/shame.js
```diff
@@ -202,3 +202,11 @@ export const shamefullyDeleteRelatedAnalysisTransforms = (graphDiv, payload) =>
...
   }
 };
+
+export const shamefullyAdjustSizeref = (gd, {update}) => {
+  const {'marker.size': size = null, 'marker.sizesrc': src = null} = update;
+  if (size && src) {
+    update['marker.sizeref'] = size.reduce((a, b) => Math.max(a, b)) / 2000;
+    update['marker.sizemode'] = 'area';
+  }
+};
```
VeraZab: well, it seems like this is a good place to put that code in.
So we can never use size.mode diameter when there's a sizeref..?

VeraZab:
this seems good, just question above..

src/shame.js
nicolaskruchten: you can, you just have to set it yourself after. It's almost always the wrong thing to do tho, so I'm OK with forcing it back every time you switch that dropdown.

PR #774, opened 2018-10-22
reuse trace type

### LINKED ISSUE #768: New trace should have same type as last trace in list
This is a smarter default behaviour than the current one, which is scatter-always.

nicolaskruchten:
Closes #768 when done...

@VeraZab @dmt0 it's actually kind of tricky to think about what the new trace should look like... should it reuse `mode`? `orientation`? Should it go through `traceTypeToPlotlyInitFigure`? Should it reuse everything except data arrays?

What did the old workspace do?

### DIFF
--- a/src/EditorControls.js
+++ b/src/EditorControls.js
@@ -1,7 +1,7 @@
 import DefaultEditor from './DefaultEditor';
 import PropTypes from 'prop-types';
 import React, {Component} from 'react';
-import {bem, localizeString} from './lib';
+import {bem, localizeString, plotlyTraceToCustomTrace, traceTypeToPlotlyInitFigure} from './lib';
 import {
   shamefullyClearAxisTypes,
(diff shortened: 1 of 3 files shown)


VeraZab:
the old workspace copied the last trace exactly, with data and everything
i think that i would have copied everything except src related info, and used `traceTypeToPlotlyInitFigure`

nicolaskruchten:
Are you sure it copied everything? I seem to recall it only copied srcs for X and not Y for scatters for example... ?

dmt0:
I'd just copy stuff that's under Structure/Traces panel, but does not involve data sources. Copying other stuff under data would have to be done on a case by case basis, and would get complicated.

dmt0:
Create area, than add one more trace -> creates line

dmt0:
💃

PR #780, opened 2018-10-30
Hot fix add trace

VeraZab:
fixes: https://github.com/plotly/streambed/issues/11752


VeraZab:
@dmt0 please review

nicolaskruchten:
:dancer:

nicolaskruchten:
How does prevtrace not have a type?? Or not exist?

nicolaskruchten:
I guess if pushed via api and relying on default type=scatter

src/lib/customTraceType.js
```diff
@@ -97,6 +97,6 @@ export function traceTypeToPlotlyInitFigure(traceType, gl = '') {
         type: 'cone',
       };
     default:
-      return {type: traceType};
+      return {type: traceType ? traceType : 'scatter'};
```
nicolaskruchten: This is actually problematic

nicolaskruchten: The scatter case should set the same options as at the top otherwise there are two “scatter” paths

nicolaskruchten:
Dancer only when scatter path is centralized :)

nicolaskruchten:
💃 looks good to me now!

PR #781, opened 2018-10-31
Trace selector props

dmt0:
Make it easier to import TraceTypeSelector - so it doesn't rely on context


src/components/widgets/TraceTypeSelector.js
```diff
@@ -243,9 +229,9 @@ TraceTypeSelector.propTypes = {
   fullValue: PropTypes.string,
   fullContainer: PropTypes.object,
   glByDefault: PropTypes.bool,
+  traceTypesConfig: PropTypes.object,
```
VeraZab: so it wasn't accessible with context after all?

VeraZab:
💃

PR #785, opened 2018-11-05
Tweaks

nicolaskruchten:
Closes https://github.com/plotly/streambed/issues/11779
Closes https://github.com/plotly/react-chart-editor/issues/562

### DIFF
--- a/src/components/fields/DataSelector.js
+++ b/src/components/fields/DataSelector.js
@@ -95,6 +95,7 @@ export class UnconnectedDataSelector extends Component {
           value={this.fullValue}
           onChange={this.updatePlot}
           multi={this.is2D}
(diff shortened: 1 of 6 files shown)


src/components/fields/DataSelector.js
```diff
@@ -95,6 +95,7 @@ export class UnconnectedDataSelector extends Component {
           value={this.fullValue}
           onChange={this.updatePlot}
           multi={this.is2D}
+          searchable={true}
```
VeraZab: as easy as that hein! nice!

VeraZab:
I think its ok that those panels are not wrapped with trace required

VeraZab:
💃

PR #787, opened 2018-11-05
use toEqual

VeraZab:


src/lib/__tests__/sortMenu-test.js
```diff
@@ -1,14 +1,12 @@
...
-function stringify(array) {
-  let string = '';
-  array.forEach(obj => {
-    string += JSON.stringify(obj);
-  });
-  return string;
-}
-
 describe('sortMenu', () => {
+  function check(actual, expected) {
```
nicolaskruchten: how does this differ from `expect(actual).toEqual(expected)` ?

VeraZab: it's 2 different arrays

PR #789, opened 2018-11-09
prevent external-fonts

VeraZab:
https://github.com/plotly/streambed/pull/11793

'Dosis' is loaded in streambed, so it's not needed here.
We could include this font locally in the editor, that doesn't seem worth it though.

Please review @nicolaskruchten @dmt0


VeraZab:
Oh 2 mins, going to choose a better default, the Percy diffs don't look right

nicolaskruchten:
💃

PR #791, opened 2018-11-11
Feature/remove context typees

oprstchn:
- change contextType component which update easily


oprstchn:
Hi @dmt0 

I have sent that PR mistakenly.

In that PR way, I can not do migration well, so I am making PR now in another way!

PR #796, opened 2018-11-14
Do not add groups and groupssrc to sort transform

### LINKED ISSUE #794: Sort transforms don't manage src very well?
@jackparmer had an issue with saving figured in Chart Studio with sort transforms.

dmt0:
Closes #794

null is not a valid src...


nicolaskruchten:
💃

PR #797, opened 2018-11-20
sortMenu adjustments

VeraZab:
Tested on streambed this time, gives desired result.
screen shot 2018-11-21 at 10 20 17 am
@nicolaskruchten @dmt0 please review

### DIFF
--- a/src/components/PanelMenuWrapper.js
+++ b/src/components/PanelMenuWrapper.js
@@ -51,11 +51,7 @@ class PanelsWithSidebar extends Component {
     const sections = [];
     const groupLookup = {};
     let groupIndex;
-    let childrenArray = React.Children.toArray(children);
-
-    if (menuPanelOrder) {
-      childrenArray = sortMenu(childrenArray, menuPanelOrder);
-    }
+    const childrenArray = sortMenu(React.Children.toArray(children), menuPanelOrder);
 
     childrenArray.forEach(child => {
       if (!child) {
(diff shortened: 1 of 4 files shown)


src/components/PanelMenuWrapper.js
```diff
@@ -51,13 +51,13 @@ class PanelsWithSidebar extends Component {
     const sections = [];
     const groupLookup = {};
     let groupIndex;
-    const panels = React.Children.toArray(children);
+    let childrenArray = React.Children.toArray(children);
```
nicolaskruchten: could be a ternary... let's try to avoid using `let` as much as we can please in new code :)

src/lib/__tests__/sortMenu-test.js
```diff
@@ -1,15 +1,19 @@
 import sortMenu from '../sortMenu';
 
 describe('sortMenu', () => {
-  it('modifies original array to follow the group, then name order provided', () => {
+  it('returns a new array of sorted children', () => {
```
nicolaskruchten: yes! much more functional :)

Modifying things in-place is almost always a bad idea

src/components/PanelMenuWrapper.js
nicolaskruchten: in fact, let's go one more step: just always call `sortMenu` and if `menuPanelOrder` is null just early-return. much easier to read.

nicolaskruchten:
💃  as-is, but if you want to do my stylistic changes that's fine too :)

VeraZab:
Tests were failing due to an intermittent CI failure, merging

PR #799, opened 2018-11-21
Feature/migration new context api

oprstchn:
- change legacy context api to new context api
- ralated plotly#776
- Because my first PR of oss, it may not follow the rules, so please comment me if I made a mistake


oprstchn:
I think it is v 16.3.0

oprstchn:
I confirmed the result of percy, but I can cure it because I have not been able to migrate perfectly yet

PR #804, opened 2018-12-11
Fix date handling

### LINKED ISSUE #803: Date axis misbehaviours
Our `Numeric` widgets don't work very well in some cases when axis-type is 'date'... In Style/Axes but also in Style/Traces like in the case of histograms with a date axis.

VeraZab:
part of #803 
- [x] ~~Adjusts AxesRange and allows to detect when fullValue is date so that proper control shows~~
- [x] Create and use NumericOrDate where applicable (there's a larger need for it)
- [x] Adjust Bin Size to take 'M[number]' format

before
screen shot 2018-12-12 at 12 43 38 pm

after
screen shot 2018-12-12 at 12 44 00 pm

### DIFF
--- a/src/components/fields/NumericOrDate.js
+++ b/src/components/fields/NumericOrDate.js
@@ -1,0 +1,38 @@
+import Field from './Field';
+import {UnconnectedNumeric} from './Numeric';
+import {UnconnectedText} from './Text';
+import PropTypes from 'prop-types';
+import React, {Component} from 'react';
+import {connectToContainer} from 'lib';
+import {isDateTime} from 'plotly.js/src/lib';
+import {isJSDate} from 'plotly.js/src/lib/dates';
+
+export class UnconnectedNumericOrDate extends Component {
+  render() {
+    const fullValueIsDate =
+      typeof this.props.fullValue === 'string' &&
+      (isDateTime(this.props.fullValue) || isJSDate(this.props.fullValue));
+
+    return fullValueIsDate ? (
+      <UnconnectedText {...this.props} placeholder={'yyyy-mm-dd 00:00:00.00'} />
+    ) : (
+      <UnconnectedNumeric {...this.props} />
+    );
+  }
+}
+
+UnconnectedNumericOrDate.propTypes = {
+  defaultValue: PropTypes.any,
+  fullValue: PropTypes.any,
+  min: PropTypes.number,
+  max: PropTypes.number,
+  multiValued: PropTypes.bool,
+  hideArrows: PropTypes.bool,
+  showSlider: PropTypes.bool,
+  step: PropTypes.number,
+  fullContainer: PropTypes.object,
+  updatePlot: PropTypes.func,
+  ...Field.propTypes,
+};
+
+export default connectToContainer(UnconnectedNumericOrDate);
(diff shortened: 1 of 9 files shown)


VeraZab:
@nicolaskruchten @dmt0 ready for review I will tackle issue 803 incrementally

src/components/fields/derived.js
```diff
@@ -196,6 +197,8 @@ export const AxesRange = connectToContainer(UnconnectedAxisRangeValue, {
   },
 });
 
+export const NumericOrDate = connectToContainer(UnconnectedNumericOrDate);
```
dmt0: Did you forget to delete this one?

VeraZab: no...
this is the connected component I'm using everywhere..

src/components/fields/NumericOrDate.js
```diff
@@ -0,0 +1,38 @@
...
+  multiValued: PropTypes.bool,
+  hideArrows: PropTypes.bool,
+  showSlider: PropTypes.bool,
+  step: PropTypes.number,
+  fullContainer: PropTypes.object,
+  updatePlot: PropTypes.func,
+  ...Field.propTypes,
+};
+
+export default connectToContainer(UnconnectedNumericOrDate);
```
dmt0: here's the other definition

VeraZab: oups.. this is the one I don't want.. thanks!

dmt0: wait! this is the one you want

dmt0:
💃 for the code

src/components/fields/derived.js
```diff
@@ -676,3 +677,15 @@ export const HoverColor = connectToContainer(UnconnectedColorPicker, {
...
 });
+
+export const BinSize = connectToContainer(UnconnectedNumeric, {
+  modifyPlotProps: (props, context, plotProps) => {
+    const {localize: _} = context;
+    if (typeof plotProps.fullValue === 'string' && plotProps.fullValue[0] === 'M') {
+      plotProps.fullValue = plotProps.fullValue.substring(1);
+      plotProps.min = 1;
+      plotProps.max = 12;
+      plotProps.units = parseInt(plotProps.fullValue, 10) === 1 ? _('Month') : _('Months');
```
nicolaskruchten: does plotly.js accept any other format than M for months? D for days? seconds? I'd be surprised if it was only months :)

nicolaskruchten:
Should we actually bake `NumericOrDate` into `AxesRange` or just have that derive from `NumericOrDate` ?

src/components/fields/derived.js
VeraZab: Well actually, from the docs, it does look that months is the only option: 
```
If the axis `type` is "date", then you must convert the time to milliseconds. For example, to set the interval between ticks to one day, set `dtick` to 86400000.0. "date" also has special values "M<n>" gives ticks spaced by a number of months. `n` must be a positive integer. To set ticks on the 15th of every third month, set `tick0` to "2000-01-15" and `dtick` to "M3". To set ticks every 4 years, set `dtick` to "M48"
```

Either that or time in milliseconds, so I could add a richer widgets that allows to do either milliseconds or months

PR #808, opened 2018-12-12
Plotly.js 1.43.0 upgrade

nicolaskruchten:
Known RCE issues:

- [x] text editor reads container[attr] and doesn't work if attr contains dots
- [x] colorscale horizontal alignment
- [x] `is2D` and array adjustments in `DataSelector`
- [x] hide hovertemplate when empty
- [x] wait for plotly.js to stop writebacks of colorscale to marker so that the layout.colorscales actually control things


nicolaskruchten:
@VeraZab if you're willing to dive back into the colorscale pickers there's a bundle of issues here although if you're sick of them maybe it's @dmt0's turn for a tour of duty in there, and you can take the others ;)

nicolaskruchten:
Let’s ignore Percy for now, I think the test failures are due to the way I pointed package.json to master. Should go away when 1.43.0 actually comes out

nicolaskruchten:
image

src/default_panels/StyleLayoutPanel.js
```diff
@@ -40,11 +40,13 @@ const StyleLayoutPanel = (props, {localize: _}) => (
     </PlotlyFold>
 
     <PlotlyFold name={_('Title')}>
-      <TextEditor attr="title.text" />
-      <FontSelector label={_('Typeface')} attr="title.font.family" clearable={false} />
-      <Numeric label={_('Font Size')} attr="title.font.size" units="px" />
-      <ColorPicker label={_('Font Color')} attr="title.font.color" />
-      <Numeric label={_('Horizontal Position')} showSlider step={0.02} attr="title.x" />
+      <PlotlySection name={_('Title')} attr="title.text">
```
dmt0: What's the point of having a section here?
image

src/default_panels/StyleTracesPanel.js
```diff
@@ -195,7 +194,38 @@ const StyleTracesPanel = (props, {localize: _}) => (
         label={_('Flatshading')}
         options={[{label: _('Enable'), value: true}, {label: _('Disable'), value: false}]}
       />
-    </PlotlySection>
+    </PlotlySection>{' '}
```
dmt0: {' '}? 🐄

dmt0:
Hover template really needs to show a list of available variables and whatever formatting options are there.

VeraZab:
In streambed, I think that we should set that watermark config option to false.. it's a bit annoying to see that plotly logo on the plot and on the header of the EditModeMenu no?

VeraZab:
I think we should find a way to make this be centered
screen shot 2018-12-17 at 11 21 31 am

VeraZab:
screen shot 2018-12-17 at 11 29 19 am

My labels and dividers are superposed, I'm not sure which option would enable me to adjust that.
And I guess I can't style the levels of axes labels individually... make first level labels smaller

Ah, I see, well with rotation it's better
screen shot 2018-12-17 at 11 33 07 am

Was just wondering if we had exposed everything related to those dividers positioning

nicolaskruchten:
that divider-positioning thing seems like a plotly.js bug, or you reset tickson to be 'labels' ? by default it's 'boundaries' for multicat, which is how this is avoided.

nicolaskruchten:
> a little more attention should be given to colorscales

say more?

VeraZab:
> The watermark config option is false by default... You want to hide the logo in streambed? It's been there forever AFAIK...

I think it just recently got restyled or something, it's a lot more visible than I remember it to be.. I don't know, just felt like overbranding. But I guess that maybe it should stay, because then that plot is saved, and wherever its used, if people don't mind it being there, its good branding for us.. Just thought it looked more visible than I remember it

VeraZab:
> a little more attention should be given to colorscales

well, maybe this:
> we could maybe remove the dropdown for the different colorscale types, as the title says that the colorscales presented to us should be of a specific type

Would make it look less jammed, since the colorscale category is in the label?

This section is more jammed than its currently on prod:
screen shot 2018-12-17 at 11 38 37 am

nicolaskruchten:
The logo was updated to match our current branding a couple of versions ago I think, and yeah, the color is punchier and no longer matches the modebar buttons so it's more visible. I think we'll keep it for now.

The watermark feature is much more invasive: when turned on, the logo is always there/doesn't fade away with the modebar. We'll likely turn it on for Community-plan embeds only.

Re colorscale pickers, I agree that the dropdown feels a bit redundant here, but people do like using sequential scales instead of categoricals even though they really shouldn't etc, so let's just keep them for now :)

nicolaskruchten:
Can you show side-by-side what's more jammed now than in prod?

VeraZab:
screen shot 2018-12-17 at 11 48 12 am

nicolaskruchten:
Ah I see. that's probably related to @dmt0's fix for alignment. although arguably there's less spacing but the scales themselves have more room. I'm OK with the current state of affairs but if we can easily add some spacing I'd be fine too.

dmt0:
💃

PR #811, opened 2018-12-17
Datetimepicker

VeraZab:
off
screen shot 2018-12-27 at 3 25 35 pm

on
screen shot 2019-01-04 at 5 34 21 pm


VeraZab:
<script></script>

VeraZab:
@nicolaskruchten ready for review : )
In the meantime, I'll look at that AM/PM indicator

src/components/fields/NumericOrDate.js
```diff
@@ -14,7 +14,7 @@ export class UnconnectedNumericOrDate extends Component {
       (isDateTime(this.props.fullValue) || isJSDate(this.props.fullValue));
 
     return fullValueIsDate ? (
-      <UnconnectedText {...this.props} placeholder={'yyyy-mm-dd 00:00:00.00'} />
+      <UnconnectedDateTimePicker {...this.props} placeholder={'yyyy-mm-dd 00:00:00.00'} />
```
nicolaskruchten: the placeholder should be `YYYY-MM-DD hh:mm:ss.xxx` or something like that, not a mix of letters and numbers :)

webpack.config.js
```diff
@@ -36,6 +36,14 @@ module.exports = {
         test: /\.(css|scss)?$/,
         use: ['style-loader', 'css-loader', 'sass-loader'],
       },
+      {
+        test: /\.(woff(2)?|ttf|eot|svg)(\?v=\d+\.\d+\.\d+)?$/,
```
nicolaskruchten: what's this for? just the little icons?

nicolaskruchten:
Let's add these to the axis range UI as well please.

nicolaskruchten:
I'm not sure what's doing this, but the dev bundle size on master is 35MB and on this branch it's 39.3MB (just from the log output of webpack) ...  is this datepicker really heavy? is fonts? there's 7k lines of CSS here in the datepicker style file...

webpack.config.js
VeraZab: yes its for the little arrow icons that come with the Datetimepicker

VeraZab:
> there's 7k lines of CSS here in the datepicker style file
yes.. i'll try to reduce those

VeraZab:
bundle size with react-day-picker, a looot better
screen shot 2019-01-04 at 12 41 48 pm

nicolaskruchten:
way better :)

VeraZab:
the original library's calendar picker looks like:
screen shot 2019-01-04 at 3 10 53 pm

changed it to:
screen shot 2019-01-04 at 3 11 25 pm

VeraZab:
Actually, I don't think I need to add the `DateTimePicker` anywhere other than in the `NumericOrDate` component.

In this pr: #804, I made sure to add NumericOrDate wherever dates were possible, so all cases should be covered, except cases where it is only possible to have dates. 

I don't see any cases where only dates are possible in the attrs currently supported in RCE.

So this one's ready for review @nicolaskruchten @dmt0 .

VeraZab:
PS: This is better reviewed by looking at the full diff because commit per commit, you'll see commits where I was using @blueprintjs/datetime, which I later replaced.

nicolaskruchten:
I'm still not seeing a datetime picker on axis range...

image

nicolaskruchten:
Some annoying behaviour:

* Date starts as `2009-12-16 12:00`
* I change the date part to `2009-12-01`
* I change the time part to something invalid like `bla` and then the whole thing switches back to `2009-12-16 12:00` (i.e. my valid date-related changes are lost)
* Note also that `24:00:00.000` which seems like it might actually be a valid time is read as invalid and resets the whole thing, meaning that the timepicker is reset to `12:00:00.000 AM` which is misleading

Not sure what we can actually do here, because I understand the mechanism (we're glueing both strings together and passing them to plotly.js, which is rejecting them and replacing them with the default, right?) but it's awkward.

VeraZab:
the DateTimePicker widget is so large actually, not because there's a date and time picker Icon along with a Text input on the same line, but because the calendar takes up a lot of space, it needs about 220px of width to fit. I made that widget of that width so that we don't see this jump when clicking on the calendar icon.



It could look like this if you prefer

VeraZab:
I think I like it the way it is, even if its not aligned with the other widgets, if there's a place where you find it really looks bad visually, maybe we could switch up the order of the widgets.

VeraZab:
Hmm, nevermind.. 
screen shot 2019-01-08 at 10 59 33 am

I guess this does look better
screen shot 2019-01-08 at 10 57 32 am

nicolaskruchten:
Yup. I prefer the jump :)

VeraZab:
here's what the change looks like
screen shot 2019-01-08 at 12 42 29 pm
screen shot 2019-01-08 at 12 42 35 pm

VeraZab:
screen shot 2019-01-08 at 4 17 32 pm

Added error validation so that only valid dates are sent to plotly.js, if not valid date or time, the border turns red, and no update is sent to plotly.js

VeraZab:
@nicolaskruchten would you play with this one a little to see if there's a user action that I might have missed?

Else I think this one's ready to merge.

nicolaskruchten:
Validation error state doesn't seem to really stop the update loop... editing any other component causes the contents to be sent anyways:

editor

VeraZab:
sorry about that @nicolaskruchten, should be ok now, was doing some cleanups and messed up the behaviour

nicolaskruchten:
Two comments about the validation:

1. not clear how to "back out" ... if I just clear the field, it resets to "yyyy-mm-dd" instead of the default
2. it doesn't seem to validate as I type, only after I hit enter or lose focus.

VeraZab:
should be ok now

VeraZab:
screen shot 2019-01-10 at 11 19 42 am

nicolaskruchten:
bingo! 💃

PR #812, opened 2018-12-17
include all trace types in constants

### LINKED ISSUE #788: Subplots panel empty for parcoords
We should minimally display a fold with some info, even if it's read-only.

Also the subplots panel should work for parcoords and sankey.

nicolaskruchten:
addresses errors from #788  ... doesn't add subplots support though!


src/lib/traceTypes.js
```diff
@@ -33,6 +29,10 @@ export const chartCategory = _ => {
       label: _('3D'),
       maxColumns: 1,
     },
+    HIDDEN: {
+      value: 'HIDDEN',
+      label: '',
```
VeraZab: so because it does not have a label its going to be hidden?

src/lib/traceTypes.js
```diff
@@ -8,10 +8,6 @@ export const chartCategory = _ => {
       value: 'SIMPLE',
       label: _('Simple'),
     },
-    CHARTS_3D: {
-      value: 'CHARTS_3D',
-      label: _('3D charts'),
-    },
```
VeraZab: no more 3d as a category?

src/lib/traceTypes.js
nicolaskruchten: no, it's hidden because it's not included in the next array

src/lib/constants.js
```diff
@@ -68,6 +68,9 @@ export const TRACE_TO_AXIS = {
     'histogram',
     'histogram2d',
     'histogram2dcontour',
+    'carpet',
+    'scattercarpet',
+    'contourcarpet',
```
VeraZab: but so, should we also add parcoords and other unsupported types, just in case someone creates them from the API, 

or I guess that just returning null in getAllAxes takes care of the panel not erroring out for those traces?

VeraZab:
seems good 💃 if that question above is how it does not error out for plots created through api but not in RCE yet

src/lib/constants.js
nicolaskruchten: returning null causes the Axes panel to be hidden :)

PR #813, opened 2018-12-17
BinSize adjustments

VeraZab:
So that it becomes this widget when axis type is date:
screen shot 2018-12-17 at 4 24 08 pm


nicolaskruchten:
I like the look! I'll let @dmt0 review the code

VeraZab:
I'm not done : )

VeraZab:
@dmt0 ready for review, would you play around with this one a little and tell me if you find anything weird (@nicolaskruchten if you have a bit of time you too )

Basically, I've chosen to convert everything to the millisecond format in bin.size to simplify conversions, but the approximations I'm making do produce some minor oddities like, going from 12 months to days gives 360 days. Also I've put upper and lower boundaries for the days and months formats. So for days, I put 366, but maybe I shouldn't, (because those could be in other calendars for ex).

nicolaskruchten:
Why do you convert the M syntax to millis? My strong instinct is to expose it, as it's a native plotly.js feature.

VeraZab:
> Why do you convert the M syntax to millis?

Added back the native 'M' support

dmt0:
Code looks fine, but it seems rather odd to me to have this specific subset of timeframes, I think users would find years useful for sure. Hours, minutes and seconds too. 

Also if we are working with months, a month really can't be assumed to be 30 days. I think it makes sense to go through the trouble of doing things moment.js way. Convert things to moment and use those functions to add months.

VeraZab:
> Years are just M12 or M24 etc.

ok given above, I've removed the `max` bounds from this component. 

I can add minutes, it will look like so:
screen shot 2018-12-18 at 9 16 08 am

nicolaskruchten:
So can we make this a dropdown and have full words instead, and then add days, minutes, seconds, milliseconds?

Also, we'll need a display rule for first-render... if I give "12345" as the value, should that show as millis or seconds or...? probably "whatever units show a log value closest to 1" will be a good rule.

VeraZab:
screen shot 2018-12-18 at 10 42 47 am
screen shot 2018-12-18 at 10 42 42 am

dmt0:
I know that a year is just 12M, but I think it's good to include it in the dropdown

nicolaskruchten:
💃

PR #815, opened 2018-12-20
Interpolate scss variable in calc()

ankoh:
The CSS still contains $colorpicker-width as Sass variables are not replaced within CSS calcs.

Sass issue: https://github.com/sass/sass/issues/818


dmt0:
💃

PR #816, opened 2018-12-20
Colorpicker in 2 fold fix

dmt0:
Colorpicker in folds within folds (as in Color Bars panel) - ensure that it doesn't resize horizontally when clicked.


nicolaskruchten:
Can we get a screenshot of before/after? there's no percy for this one.

dmt0:
before: 
image
after:
image

nicolaskruchten:
💃

Issue #817, opened 2018-12-21
Heatmap chart - colorscale not changing in Style/Traces


priyatharsan:

priyatharsan:
Here is the workflow: 
![heatmap-colorscale-issue](https://user-images.githubusercontent.com/25846124/50355402-96477780-051c-11e9-835a-b2b60603ee5a.gif)

PR #818, opened 2018-12-21
add autocolorscale false when user updates colorscale

VeraZab:
closes: #817


VeraZab:
I've just tried this locally, and Style/Traces works, it works for multiple traces as well (ex 2 histogram traces, styling both from the 'by type' panel, or 'individually'

src/components/fields/ColorscalePicker.js
```diff
@@ -22,6 +23,13 @@ export class UnconnectedColorscalePicker extends Component {
         }),
         colorscaleType
       );
+      this.context.onUpdate({
```
nicolaskruchten: can't reuse the existing update call somehow?

nicolaskruchten:
💃

Issue #822, opened 2019-01-08
Validate Numerics


nicolaskruchten:
Add #811-style validation to numerics! should be reasonably easy and magically adds validation to like 90% of the RCE :)

PR #823, opened 2019-01-09
add error state when !isNumber to Numeric component

VeraZab:
closes: #822 

screen shot 2019-01-09 at 11 15 41 am

same style as in this pr:
https://github.com/plotly/react-chart-editor/pull/811#issuecomment-452454940

### DIFF
--- a/src/components/widgets/NumericInput.js
+++ b/src/components/widgets/NumericInput.js
@@ -12,11 +12,23 @@ export default class NumericInput extends Component {
   constructor(props) {
     super(props);
 
-    this.state = {value: props.value};
+    this.state = {
+      value: props.value,
+      numericInputClassName: this.getNumericInputClassName(props.value),
+    };
+
     this.onChange = this.onChange.bind(this);
     this.updateValue = this.updateValue.bind(this);
     this.onKeyDown = this.onKeyDown.bind(this);
     this.onWheel = this.onWheel.bind(this);
+  }
+
+  getNumericInputClassName(value) {
+    return isNumeric(value)
+      ? `numeric-input__number ${this.props.editableClassName ? this.props.editableClassName : ''}`
+      : `numeric-input__number--error ${
+          this.props.editableClassName ? this.props.editableClassName : ''
+        }`;
   }
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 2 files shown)


VeraZab:
Because NumericInput is the base widget used in all numerics (DualNumeric, derived components) adding validation to it should be working for anywhere there's numerics involved.

@dmt0 @nicolaskruchten ready for review

src/styles/components/widgets/_numeric-input.scss
```diff
@@ -49,6 +49,26 @@
...
+  text-align: left;
+  border-radius: var(--border-radius-small);
+  padding: var(--spacing-quarter-unit) var(--spacing-quarter-unit) var(--spacing-quarter-unit)
+    var(--spacing-half-unit);
+  width: 62px;
+  vertical-align: middle;
+  font-size: inherit;
+  color: inherit;
+  font-family: inherit;
+}
```
dmt0: We don't need to copy the whole class over, do we? We can just add a class with the difference.

src/components/widgets/NumericInput.js
```diff
@@ -12,13 +12,25 @@ export default class NumericInput extends Component {
   constructor(props) {
     super(props);
 
-    this.state = {value: props.value};
+    this.state = {
+      value: props.value,
```
dmt0: I know this is not a new code addition, but assigning props to state is such an antipattern.

src/styles/components/widgets/_numeric-input.scss
VeraZab: yeah, I could do `@extend .numeric-input__number;`

src/components/widgets/NumericInput.js
VeraZab: that's our way of not having plotly.js reject the value and override with its default

nicolaskruchten:
On my screen the little gray fuzzy focus style hides the red... 

image

VeraZab:
focussed version:
screen shot 2019-01-09 at 2 17 31 pm

VeraZab:
Can I merge this one?

nicolaskruchten:
same comments as the other validation one: there's no easy "back out to defaults" mode... once it's red it stays red and deleting it doesn't undo... Can we have it accept `""` such that plotly.js will then go back to defaults?

nicolaskruchten:
:dancer:

Issue #825, opened 2019-01-09
setting colorscale to [] breaks plotly.py


nicolaskruchten:
https://community.plot.ly/t/chart-studio-get-plotly-version-for-python-export/17882

nicolaskruchten:
IIRC we did this in a couple of places to 'reset' the colorscale, but we should be setting it to null or something instead.

PR #826, opened 2019-01-11
change colorscale reset value: [] => null (for python api)

VeraZab:
closes: #825

I tested this and it does not seem like it breaks anything for us to have the reset value, when colorscale is a variable, to be null instead of empty array.

I tested this in RCE and streambed:
- tested switching marker.color from constant to variable and changing scales, for individual and grouped traces
- tested making pie charts and changing their colors

### DIFF
--- a/src/components/fields/MarkerColor.js
+++ b/src/components/fields/MarkerColor.js
@@ -58,7 +58,7 @@ class UnconnectedMarkerColor extends Component {
         this.context.updateContainer({
           ['marker.color']: null,
           ['marker.colorsrc']: null,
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
After a bunch of adding and removing of colorscales, I got to this state:
image

nicolaskruchten:
Basically I no longer have the ability to choose the colorscale, and there is none. I'll try to provide a clear replication path

nicolaskruchten:
can't replicate :(

nicolaskruchten:
let's merge and release once this clears tests

PR #846, opened 2019-02-17
Fix meta titles

VeraZab:
closes: https://github.com/plotly/streambed/issues/12502


VeraZab:
note that setting trace titles, timescale buttons, is not yet possible, but will be in next plotly.js release: https://github.com/plotly/streambed/issues/12434

bpostlethwaite:
Any other places we need the template string? @dmt0 care to take a look? If not, let's get this in and do another PR later if necessary

PR #853, opened 2019-02-25
Package updates

### LINKED ISSUE #245: upgrade babel to @babel v7
online documentation is slowly being migrated to the new `@babel` way of doing things. May as well get this one :lock: away.

### LINKED ISSUE #585: Migrate from percy react to storybook
Ugh.

dmt0:
Closes #245 
Closes #585


dmt0:
@VeraZab @nicolaskruchten Ready for review commit by commit

src/components/containers/ModalProvider.js
```diff
@@ -73,10 +73,10 @@ class ModalProvider extends React.Component {
   render() {
     const {component: Component, componentProps, isAnimatingOut} = this.state;
     return (
-      <Fragment>
+      <>
```
VeraZab: oh wow..

src/EditorControls.js
```diff
@@ -40,8 +40,6 @@ class EditorControls extends Component {
       data: gd.data,
       dataSources: this.props.dataSources,
       dataSourceOptions: this.props.dataSourceOptions,
-      dataSourceValueRenderer: this.props.dataSourceValueRenderer,
```
VeraZab: euh...why are we removing this?
this is actually a prop that's needed for one of our clients..

VeraZab: we can't just delete props we used to have on our components, streambed is not the only RCE user

src/components/fields/DataSelector.js
```diff
@@ -117,8 +117,6 @@ export class UnconnectedDataSelector extends Component {
           onChange={this.updatePlot}
           multi={this.is2D}
           searchable={true}
-          optionRenderer={this.context.dataSourceOptionRenderer}
```
VeraZab: same thing here, this is actually needed by a client

src/components/fields/Dropdown.js
```diff
@@ -33,10 +31,8 @@ UnconnectedDropdown.propTypes = {
   backgroundDark: PropTypes.bool,
   clearable: PropTypes.bool,
   fullValue: PropTypes.any,
-  optionRenderer: PropTypes.func,
   options: PropTypes.array.isRequired,
   updatePlot: PropTypes.func,
-  valueRenderer: PropTypes.func,
```
VeraZab: same

src/components/fields/DropdownCustom.js
```diff
@@ -59,8 +59,6 @@ export class UnconnectedDropdownCustom extends Component {
           value={value}
           onChange={this.setValue}
           clearable={this.props.clearable}
-          optionRenderer={this.props.optionRenderer}
-          valueRenderer={this.props.valueRenderer}
```
VeraZab: same (needed)

src/components/containers/ModalProvider.js
dmt0: yep, that's about the only benefit we get from babel 7 😛

src/EditorControls.js
dmt0: Oh, we have to talk...

src/components/fields/FontSelector.js
```diff
@@ -2,29 +2,21 @@ import Dropdown from './Dropdown';
...
+    }))}
+    clearable={true}
+  />
+);
 
 FontSelector.propTypes = {
   ...Dropdown.propTypes,
 };
 
-FontSelector.defaultProps = {
```
VeraZab: shouldn't we keep the default value we had? and let clearable be an option we set through props..

src/components/fields/LineSelectors.js
```diff
@@ -93,13 +82,7 @@ class LineSelector extends Component {
 
   render() {
     return (
-      <Dropdown
-        {...this.props}
-        options={this.options}
-        valueRenderer={styledRenderer}
```
VeraZab: again.. needed

src/EditorControls.js
dmt0: Actually, as long as we don't hit 1.0, we can change the API in any way we want!

src/components/widgets/Dropdown.js
```diff
@@ -48,35 +43,33 @@ class Dropdown extends Component {
...
+          isClearable={clearable}
+          value={opts.filter(o =>
+            Array.isArray(value) ? value.includes(o.value) : value === o.value
+          )}
           options={opts}
-          searchable={searchable}
+          isSearchable={searchable}
           onChange={this.onChange}
-          multi={multi}
-          optionRenderer={optionRenderer}
```
VeraZab: here again

src/components/widgets/Dropdown.js
```diff
@@ -48,35 +43,33 @@ class Dropdown extends Component {
...
     const dropdownContainerClass = classnames('dropdown-container', {
       'dropdown--dark': this.props.backgroundDark,
-      [this.props.className]: this.props.className,
+      [className]: className,
     });
 
     return (
       <div className={dropdownContainerClass} style={dropdownStyle}>
         <Select
-          backspaceToRemoveMessage={backspaceToRemoveMessage}
```
VeraZab: this prop does not exist anymore?

src/components/widgets/Dropdown.js
```diff
@@ -48,35 +43,33 @@ class Dropdown extends Component {
...
+          )}
           options={opts}
-          searchable={searchable}
+          isSearchable={searchable}
           onChange={this.onChange}
-          multi={multi}
-          optionRenderer={optionRenderer}
-          valueRenderer={valueRenderer}
-          noResultsText={noResultsText || _('No Results')}
-          valueKey={valueKey}
```
VeraZab: how about this, not needed anymore?

src/styles/components/widgets/_dropdown.scss
```diff
@@ -1,127 +1,121 @@
 .dropdown-container {
-  @import 'node_modules/react-select/scss/default';
+  // @import 'node_modules/react-select/scss/default';
```
VeraZab: not needed?

src/lib/striptags.js
```diff
@@ -24,223 +24,210 @@
 
 'use strict';
 
-(function(global) {
```
VeraZab: this is copy pasted from somewhere?

.eslintrc
```diff
@@ -60,7 +60,7 @@
     "guard-for-in": ["off"],
     "import/named": ["off"],
     "import/no-duplicates": ["error"],
-    "import/no-named-as-default": ["error"],
+    "import/no-named-as-default": ["off"],
```
VeraZab: why do we want to turn this off?

README.md
```diff
@@ -155,14 +155,14 @@ Once you have your tokens, you can provide it as a config prop to the `<PlotlyEd
...
 
-&copy; 2018 Plotly, Inc. MIT License.
+&copy; 2019 Plotly, Inc. MIT License.
 
 ## Sponsors
 
 This work was made possible by [Alteryx](https://www.alteryx.com/), a leader in self-serve data analytics and data blending.
 
-![alteryx](https://www.alteryx.com/sites/all/themes/alteryx2014/_media/alteryx-logo.png)
+![alteryx](https://www.alteryx.com/inspire-europe-2018/photos_files/alteryx-logo.png)
```
VeraZab: 👍 :)

src/EditorControls.js
VeraZab: well.. no im not sure.. let's talk

src/components/containers/ModalProvider.js
VeraZab: lol, I kind of liked it better when it was called Fragment and we saw where it was imported from, just thinking of devs new to React who see all those weird little syntaxes, that show little clue as to where they're coming from..
but oh well.. that's how it is I guess

VeraZab:
yeah, mainly have a concern about those `optionRenderer`, and `valueRenderer` props we've removed.. 

@bpostlethwaite, would be nice to have your input here, maybe we could have a quick Slack chat about this, do you think its alright to remove those 2 props from our api..?

src/components/fields/LineSelectors.js
dmt0: This particular spot - do we really need them here? It's a line selector, it's not a data source thing, we don't pass things here down from EditorControls. I don't think we need these props here.

src/components/widgets/Dropdown.js
dmt0: noResultsText = noOptionsMessage
The rest are superseded by `components`

.eslintrc
dmt0: It looks like this has never been enforced, but now it is, probably because of webpack 4 upgrade. Here's the rule:
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-named-as-default.md
We have this all over the place, our every export is named, and then we import it by that name. The rule seems reasonable, but it's too much work to follow it now.

src/lib/striptags.js
dmt0: That piece of code aims to support different module systems. Our setup started to complain about it, so I simplified it. No functionality lost.

src/components/widgets/Dropdown.js
dmt0: > may be implemented in a later version
:(
not a critical prop though

dmt0:
Regarding the removed props. They were now replaced in `EditorControls` with this:
`  dataSourceComponents: PropTypes.object,`
It gives us the same functionality, through a different mechanism though.
The props are not mentioned in the readme, but our version bump should include something like this:

**Breaking change:** 
`dataSourceValueRenderer` and `dataSourceOptionRenderer` props in `EditorControls` have been replaced with a new prop `dataSourceComponents`. See the upgrade guide here: https://react-select.com/upgrade-guide#new-components-api

bpostlethwaite:
Yes we can remove those from our API, documented in the Changelog of course. Thanks for bringing this to my attention!

src/styles/components/widgets/_dropdown.scss
```diff
@@ -1,127 +1,120 @@
...
-      & > .Select-control {
-        background-color: var(--color-background-inputs);
-        .Select-input {
-          background-color: var(--color-background-inputs) !important;
-        }
-      }
-    }
-    &.is-open {
-      & > .Select-control {
+    &__control {
```
VeraZab: aie, I think there's going to be streambed css weirdness in this one... 
I think I remember having to style some of this in a special way, we'll see..

VeraZab:
ok, well 💃 from me!

src/styles/components/widgets/_dropdown.scss
dmt0: I expect to remove a bunch of "integration" stuff in streambed after this, as it won't be necessary anymore. Especially multi-select related stuff. Won't be a trivial update.

Issue #873, opened 2019-03-13
scattermapbox SymbolSelector to accept a string


dmt0:
The following is a spec for `marker.symbol` for `scattermapbox` trace:

Type: (string or array of strings)
Default: circle
Sets the marker symbol. Full list: https://www.mapbox.com/maki-icons/ Note that the array marker.color and marker.size are only available for circle symbols.

Currently we only have a dropbox with predefined symbols

PR #874, opened 2019-03-13
Supported symbols in SymbolSelector

### LINKED ISSUE #857: [3D] marker symbols are limited in plotly.js for 3D but RCE shows all the available marker symbols in the dropdown that won't support
Supported 3D marker symbols are: https://github.com/plotly/plotly.js/blob/master/src/constants/gl3d_markers.js

3d_marker_symbols

dmt0:
Closes #857
Note: This field should be a text input for `scattermapbox` as it supports maki icons. This is not being implemented in this PR.

Type: (string or array of strings)
Default: circle
Sets the marker symbol. Full list: https://www.mapbox.com/maki-icons/ Note that the array marker.color and marker.size are only available for circle symbols.

New issue is raised for that: https://github.com/plotly/react-chart-editor/issues/873


dmt0:
@VeraZab ready for review

src/components/fields/SymbolSelector.js
```diff
@@ -377,10 +377,10 @@ class SymbolSelector extends Component {
...
     this.borderColor = Array.isArray(this.borderColor) ? COLORS.charcoal : this.borderColor;
 
-    if (this.props.is3D) {
-      this.symbolOptions = SYMBOLS.filter(option => {
-        return option.threeD;
-      });
+    if (this.props.container.type === 'scatter3d') {
+      this.symbolOptions = SYMBOLS.filter(option => option.threeD);
+    } else if (this.props.container.type === 'scattermapbox') {
+      this.symbolOptions = [SYMBOLS[0]];
```
VeraZab: So when its scattermapbox, we only allow for the circle symbol? should we not allow at least some of the other variations, like open circle, etc?

dmt0: Yeah, actually all of them are supported alright, no need to treat it as special case. Fixed.

VeraZab:
💃

PR #884, opened 2019-03-25
Dropdown fixes

dmt0:
- Fixes situation when dropdown has options with value field other than 'value'
- Fixes transparent background of dropdown menu when there are no options


VeraZab:
💃

PR #893, opened 2019-04-05
Sunburst

VeraZab:
left out the sunburst transforms for now


src/components/fields/derived.js
```diff
@@ -161,7 +161,10 @@ export const TickFormat = connectToContainer(UnconnectedDropdownCustom, {
 
 export const ShowInLegend = connectToContainer(UnconnectedVisibilitySelect, {
   modifyPlotProps: (props, context, plotProps) => {
-    plotProps.isVisible = context.fullLayout.showlegend;
+    if (context.type && context.type !== sunburst) {
```
nicolaskruchten: please add choropleth here?

src/components/fields/derived.js
```diff
@@ -588,6 +591,14 @@ export const HoverInfo = connectToContainer(UnconnectedFlaglist, {
         {label: _('Norm'), value: 'norm'},
         {label: _('Divergence'), value: 'divergence'},
       ];
+    } else if (container.type === 'sunburst') {
+      options = [];
+      if (container.labels) {
```
nicolaskruchten: can we align this with `pie` above plz? in pie we don't do the container checks but should we?

src/default_panels/GraphCreatePanel.js
```diff
@@ -30,8 +30,25 @@ const GraphCreatePanel = (props, {localize: _, setPanel}) => {
 
       <LocationSelector attr="type" />
 
-      <DataSelector label={_('Values')} attr="values" />
-      <DataSelector label={_('Labels')} attr="labels" />
+      <TraceTypeSection traceTypes={['pie']} mode="trace">
+        <DataSelector label={_('Values')} attr="values" />
```
nicolaskruchten: why isolate here instead of reuse?

src/default_panels/StyleTracesPanel.js
```diff
@@ -42,12 +42,18 @@ import {
...
   TextInfo,
   HoveronDropdown,
+  LevelRendered,
 } from '../components/fields/derived';
 
 const StyleTracesPanel = (props, {localize: _}) => (
   <TraceAccordion canGroup>
     <TextEditor label={_('Name')} attr="name" richTextOnly />
     <NumericFraction label={_('Trace Opacity')} attr="opacity" />
+    <TraceTypeSection name={_('Leafs')} traceTypes={['sunburst']} mode="trace">
```
nicolaskruchten: Leaves.

src/components/fields/derived.js
VeraZab: choropleth doesn't have that Legend Section showing up

src/default_panels/GraphCreatePanel.js
VeraZab: It seemed to me that the preferred order in which these show up in Pie vs Sunburst traces would be different..

I can remove if you don't mind the order

VeraZab:
@dmt0 ready for code review please

src/components/containers/SubplotAccordion.js
```diff
@@ -113,25 +114,28 @@ class SubplotAccordion extends Component {
...
-    let tableCounter = 0;
+    const counter = {
+      pie: 0,
+      table: 0,
+      sunburst: 0,
+    };
+
     data.forEach((d, i) => {
-      if ((d.type === 'pie' && d.values) || d.type === 'table') {
+      if ((d.type === 'pie' && d.values) || d.type === 'table' || d.type === 'sunburst') {
```
dmt0: Do we need this outer condition? Seems like double work. Perhaps it can be removed? Just have to make sure we still check for `d.values` in `pie` case.

src/components/containers/SubplotAccordion.js
```diff
@@ -113,25 +114,28 @@ class SubplotAccordion extends Component {
...
           <TraceFold
             key={i}
             traceIndexes={[i]}
             canDelete={false}
-            name={
-              d.type === 'pie'
-                ? `${_('Pie')} ${pieCounter > 1 ? pieCounter : ''}`
-                : `${_('Table')} ${tableCounter > 1 ? tableCounter : ''}`
-            }
+            name={capitalize(d.type) + ' ' + counter[d.type]}
```
dmt0: You left out the `Counter > 1` condition, you still need it...

VeraZab: i think its ok, if the counter is at zero, that trace does not even exist in data, so there won't be an accordion for it..

unless im missing something else?

VeraZab: ah right nevermind, it can be 1

dmt0: The condition says that if you only have 1 trace, don't include number in the subplot fold name. So you don't get "Pie 1", it'll be just "Pie". Numbers kick in only if you have more than 1. So you do need that condition, otherwise it's not pretty.

src/components/containers/SubplotAccordion.js
```diff
@@ -122,13 +122,7 @@ class SubplotAccordion extends Component {
...
     data.forEach((d, i) => {
       if ((d.type === 'pie' && d.values) || d.type === 'table' || d.type === 'sunburst') {
-        if (d.type === 'pie') {
-          counter.pie++;
-        } else if (d.type === 'table') {
-          counter.table++;
-        } else if (d.type === 'sunburst') {
-          counter.sunburst++;
-        }
+        counter[d.type]++;
```
dmt0: 👍

dmt0:
💃 !

VeraZab:
@nicolaskruchten ready for your final ✨ approval

src/default_panels/GraphCreatePanel.js
nicolaskruchten: I don't mind the order

src/default_panels/StyleTracesPanel.js
```diff
@@ -42,12 +42,18 @@ import {
...
   HoveronDropdown,
+  LevelRendered,
 } from '../components/fields/derived';
 
 const StyleTracesPanel = (props, {localize: _}) => (
   <TraceAccordion canGroup>
     <TextEditor label={_('Name')} attr="name" richTextOnly />
     <NumericFraction label={_('Trace Opacity')} attr="opacity" />
+    <TraceTypeSection name={_('Leaves')} traceTypes={['sunburst']} mode="trace">
+      <LevelRendered label={_('Start at Level')} attr="level" />
```
nicolaskruchten: wow this is nicely done!

nicolaskruchten:
✨ this looks great :)

PR #898, opened 2019-04-08
Sankey

### LINKED ISSUE #788: Subplots panel empty for parcoords

### LINKED ISSUE #896: dropdown CSS breakage
In the demo app today, when selecting a color scale:

image

dmt0:
Closes #788
Closes #896
Implements Sankey trace type


dmt0:
@nicolaskruchten @VeraZab Ready for review
I'm open to a better name for `ColorArrayPicker` - which is just a switch between a constant and an array color.

VeraZab:
do we need to add a sankey.json for percy tests?

dev/percy/index.js
```diff
@@ -1,11 +1,10 @@
...
+export {default as panelTest} from './panelTest.json';
+export {default as histogram} from './histogram.json';
+export {default as histogram2d} from './histogram2d.json';
+export {default as pie} from './pie.json';
+export {default as violin} from './violin.json';
+export {default as bar} from './bar.json';
+export {default as box} from './box.json';
+export {default as waterfall} from './waterfall.json';
+export {default as sunburst} from './sunburst.json';
+export {default as sankey} from './sankey.json';
```
dmt0: I think this syntax is easier, since you don't need to add something in two places. What do you guys think? Shall we put it in more places?
And no, Vera, you can't just do `export sankey from...`

VeraZab: yeah, I know :) that's because the sankey.json file doesn't have an export default = ...

dmt0: No, actually a default export is all there is, just syntax doesn't seem to allow and 'export from' that way...

nicolaskruchten: Can you export * from ... ?

dmt0: Well, in case of JSON they won't be named...

VeraZab:
its weird, when I load a sankey, from the `dev/json` mock search: percy/sankey.json, I get:
Screen Shot 2019-04-10 at 2 06 27 PM

The data selectors, don't show that data is inlined, but also no datasrcs appear in the panel

dmt0:
Yeah, I use inlined data and then data source names from test fixtures. That way the dropdowns usually get filled in percy. But yes, it's a hack...

nicolaskruchten:
yeah the inline thing seems to be brittle in ways that I don't fully understand :(

src/default_panels/StyleTracesPanel.js
```diff
@@ -694,6 +696,29 @@ const StyleTracesPanel = (props, {localize: _}) => (
...
+          {label: _('Snap'), value: 'snap'},
+          {label: _('Perpendicular'), value: 'perpendicular'},
+          {label: _('Freeform'), value: 'freeform'},
+          {label: _('Fixed'), value: 'fixed'},
+        ]}
+        clearable={false}
+      />
+    </PlotlySection>
+    <PlotlySection name={_('Links')}>
+      <ColorArrayPicker label={_('Color')} attr="link.color" rootAttr="link" />
```
VeraZab: rootAttr, that can be infered, from attr no?
like attr.split('.')[0]?

src/components/fields/ColorArrayPicker.js
```diff
@@ -0,0 +1,133 @@
+import Field from './Field';
```
VeraZab: actually, I think this one is valid to put in a whole bunch of more places, basically anywhere where we have the MultiColorPicker, if should actually be this one, to allow to color by array, its an ability we didn't add before, and some clients have asked for it 

cc @nicolaskruchten : what do you think, should we make this generalizable and apply more widely?

nicolaskruchten:
✨ once one change is made: please move links.labels to below links.values. My reasoning is that the absolute minimal data input for Sankey is `link.{sources,targets,values}` ... surprisingly nothing from `node` is strictly required (although we should keep `node` above `link` as it is now IMO)

dmt0:
Are you guys OK with the name ColorArrayPicker? We have too many color picker, I'm out of names...

src/components/fields/ColorArrayPicker.js
nicolaskruchten: you're right that it could apply more widely but i hesitate a bit... it'll make the whole UI even heavier than it already is if we have all these constant-vs-variable toggles everywhere. can we create an issue and just do it here for now?

I think I'd like to have a smaller set of toggles for these conditional-widgets.

Issue #717, opened 2018-08-29
Reorderable traces


nicolaskruchten:
Would be nice to (again) complexify the trace accordion to allow for reordering of traces via drag and drop.

nicolaskruchten:
Let's just add little up/down chevrons to the Fold titlebar next to the X in Structure/Traces *only* ... not in the other trace accordions

PR #902, opened 2019-04-09
Add ability to reorder folds

VeraZab:
closes: #717

### DIFF
--- a/src/EditorControls.js
+++ b/src/EditorControls.js
@@ -290,8 +290,8 @@ class EditorControls extends Component {
         break;
 
       case EDITOR_ACTIONS.MOVE_TO:
-        // checking is fromIndex and toIndex is a number because just checking
-        // if not there will not work as index can be 0 and that value is falsy
+        // checking if fromIndex and toIndex is a number because
+        // gives errors if index is 0 (falsy value)
         if (payload.path && !isNaN(payload.fromIndex) && !isNaN(payload.toIndex)) {
           function move(container) {
             const movedEl = container[payload.fromIndex];
--- a/src/EditorControls.js
+++ b/src/EditorControls.js
@@ -293,18 +293,23 @@ class EditorControls extends Component {
         // checking is fromIndex and toIndex is a number because just checking
         // if not there will not work as index can be 0 and that value is falsy
         if (payload.path && !isNaN(payload.fromIndex) && !isNaN(payload.toIndex)) {
+          function move(container) {
+            const movedEl = container[payload.fromIndex];
+            const replacedEl = container[payload.toIndex];
+            container[payload.toIndex] = movedEl;
+            container[payload.fromIndex] = replacedEl;
+          }
+
           if (payload.path === 'data') {
-            const traceMoved = graphDiv.data[payload.fromIndex];
-            const traceReplaced = graphDiv.data[payload.toIndex];
-            graphDiv.data[payload.toIndex] = traceMoved;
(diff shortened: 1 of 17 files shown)


VeraZab:
@nicolaskruchten I think this is what we want would you check please ✨ 
I didn't add it to other accordions, but this easily could be extended to them

VeraZab:
@dmt0 I think this is ready for code review as well

VeraZab:
adjusted styles
Screen Shot 2019-04-10 at 11 02 25 AM

VeraZab:
so I've added the ability to reorder: images, shapes, annotations on top of traces

for transforms, I wasn't sure about the array to reorder (it changes based on the transform applied, and I'm not sure of the side effects) so I didn't do it here.

for rangeselector buttons, the path to get to the array to reorder is a bit less straight forward. I could perhaps use the `nested property` function of plotly.js. but didn't do it here.

just picked the low hanging fruit. let me know if that's ok @nicolaskruchten

src/components/containers/PlotlyFold.js
```diff
@@ -72,8 +72,30 @@ export class Fold extends Component {
...
+      <div className="fold__top__moving-controls">
+        <span
+          className="fold__top__moving-controls--up"
+          onClick={e => {
+            // prevents fold toggle to happen when clicking on moving arrow controls
+            e.stopPropagation();
+            alert('up');
+          }}
+        >
+          {canMoveUp ? <AngleDownIcon /> : null}
```
dmt0: would it make sense to move this check outside the `span`?

src/EditorControls.js
```diff
@@ -289,6 +289,26 @@ class EditorControls extends Component {
         }
         break;
 
+      case EDITOR_ACTIONS.MOVE_TO:
+        // checking is fromIndex and toIndex is a number because just checking
+        // if not there will not work as index can be 0 and that value is falsy
```
dmt0: very hard to read this comment :)

src/EditorControls.js
```diff
@@ -294,10 +294,10 @@ class EditorControls extends Component {
         // if not there will not work as index can be 0 and that value is falsy
         if (payload.path && !isNaN(payload.fromIndex) && !isNaN(payload.toIndex)) {
           if (payload.path === 'data') {
-            const a = graphDiv.data[payload.fromIndex];
-            const b = graphDiv.data[payload.toIndex];
-            graphDiv.data.splice(payload.toIndex, 1, a);
-            graphDiv.data.splice(payload.fromIndex, 1, b);
+            const traceMoved = graphDiv.data[payload.fromIndex];
+            const traceReplaced = graphDiv.data[payload.toIndex];
```
dmt0: `traceFrom`, `traceTo`? When you flip two traces, how do you decide which one is moved and which one is replaced?

src/components/containers/PlotlyFold.js
dmt0: nevermind...

dmt0:
💃 !

src/EditorControls.js
VeraZab: yup sorry, should have said that this one is not reviewable commit by commit..

nicolaskruchten:
✨ although I don't love the "you can't click here" cursor on the greyed out arrows... I'd just leave it as a normal non-link cursor.

VeraZab:
> although I don't love the "you can't click here" cursor on the greyed out arrows
yeah, agree, removed!

PR #904, opened 2019-04-11
StyleGeoPanel

### LINKED ISSUE #851: geo layout attributes only visible in subplots tab
It took me a pretty long time/a few tries to find the `geo` attributes (`projection`, map options, ) when creating a map, it looks like those options are currently available in the Subplot tab:
screen shot 2019-02-20 at 4 57 17 pm

I'm not sure if this is a bug or was intended but can we consider adding these features to one of the style tabs, I don't think it makes sense in the subplots section as users won't visit that section when plotting only a single trace

cc @VeraZab @dmt0

dmt0:
Closes #851
Also hide Legend if there are no traces with `showlegend` defined in `fullData`


nicolaskruchten:
let's get Percy up so I can see what this looks like! I'd like to see the Geo panel in percy if possible, with a mock that has 2 geos and a scatter... I'd expect to see just two subplot folds, right? one for each geo, but the scatter should be filtered out (which i'm not seeing any code for but maybe I just missed)

src/DefaultEditor.js
```diff
@@ -64,6 +66,18 @@ class DefaultEditor extends Component {
     return this.context.fullData.some(d => traceHasColorbar({}, d));
   }
 
+  hasLegend() {
+    return this.context.fullData.some(t => t.showlegend !== undefined); // eslint-disable-line no-undefined
```
VeraZab: are we sure that this still enables us to press the 'showlegend' button, I think that in some cases its hidden by default no?

src/default_panels/StyleMapsPanel.js
```diff
@@ -0,0 +1,167 @@
...
+  SubplotAccordion,
+  PlotlySection,
+  Dropdown,
+  Radio,
+  Numeric,
+  ColorPicker,
+} from '../components';
+
+const StyleMapsPanel = (props, {localize: _}) => (
+  <SubplotAccordion>
```
VeraZab: 👍

src/default_panels/index.js
```diff
@@ -1,29 +1,14 @@
...
-  StyleNotesPanel,
-  StyleShapesPanel,
-  StyleSlidersPanel,
-  StyleImagesPanel,
-  StyleTracesPanel,
-  StyleColorbarsPanel,
-  StyleUpdateMenusPanel,
-  GraphSubplotsPanel,
-};
+export {default as GraphCreatePanel} from './GraphCreatePanel';
```
VeraZab: 👍

VeraZab:
yeah, just wondering if that Legend behavior is good, otherwise 💃

src/DefaultEditor.js
dmt0: This hides the panel only if there are absolutely no traces that support `showlegend`. So all should be fine.

PR #934, opened 2019-06-11
Funnel support

nicolaskruchten:
Outstanding:

- [x] y autorange for horizontal funnels
- [x] barmode
- [x] class as function
- [x] hover/textinfo on funnel
- [x] Percy

### DIFF
--- a/dev/percy/funnel.json
+++ b/dev/percy/funnel.json
@@ -1,0 +1,9 @@
+{
+  "data": [
+    {
+      "type": "funnel",
+      "y": ["Lead", "Pipeline", "Proposal", "Negotiation", "Closed (Won)"],
+      "x": [ 610, 432, 231, 103, 54 ]
+    }
+  ]
+}
--- a/src/components/containers/SubplotAccordion.js
+++ b/src/components/containers/SubplotAccordion.js
@@ -64,7 +64,7 @@ class SubplotAccordion extends Component {
     // of right type that have attr 'subplot': 'ternary' in their data.
 
     /**
-    Example: 
+    Example:
     {
       "data": [
         {
@@ -126,7 +126,9 @@ class SubplotAccordion extends Component {
     data.forEach((d, i) => {
       if (
         (d.type === 'pie' && d.values) ||
-        ['pie', 'table', 'sunburst', 'sankey', 'parcoords', 'parcats'].includes(d.type)
+        ['pie', 'table', 'sunburst', 'sankey', 'parcoords', 'parcats', 'funnelarea'].includes(
+          d.type
+        )
       ) {
         counter[d.type]++;
         const currentCount = counter[d.type];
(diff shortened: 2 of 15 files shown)


dev/percy/funnel.json
```diff
@@ -0,0 +1,9 @@
+{
+  "data": [
+    {
+      "type": "funnel",
+      "y": ["Lead", "Pipeline", "Proposal", "Negotiation", "Closed (Won)"],
```
VeraZab: just a general, more plot.js type question. Why are the axes for funnel (x/y) and funnelarea (labels/values), why couldn't they both have been named (labels/values) ? I feel like that would have been easier understand for a business type chart.

src/components/containers/SubplotAccordion.js
```diff
@@ -126,7 +126,9 @@ class SubplotAccordion extends Component {
     data.forEach((d, i) => {
       if (
         (d.type === 'pie' && d.values) ||
-        ['pie', 'table', 'sunburst', 'sankey', 'parcoords', 'parcats'].includes(d.type)
+        ['pie', 'table', 'sunburst', 'sankey', 'parcoords', 'parcats', 'funnelarea'].includes(
+          d.type
+        )
```
VeraZab: it's nice that we can make subplots with funnelarea, but the legends will overlap, unless there's a control i've missed for that?
Screen Shot 2019-06-13 at 10 51 20 AM

src/lib/customTraceType.js
```diff
@@ -45,29 +43,10 @@ export function traceTypeToPlotlyInitFigure(traceType, gl = '') {
...
-        type: 'candlestick',
-        decreasing: {
-          line: {color: COLORS.middleGray},
-          fillcolor: 'rgba(127, 127, 127, 0.5)',
-        },
-        increasing: {
-          line: {color: COLORS.blueTeal},
-          fillcolor: 'rgba(23, 190, 207, 0.5)',
-        },
+        orientation: 'v',
```
VeraZab: well, I was asked for those colors previously.. but if we don't need them anymore ok..

VeraZab:
well, overall, this looks good to me 💃

dev/percy/funnel.json
nicolaskruchten: yeah this is because `funnel` is based on `bar` and is on cartesian axes, but `funnelarea` is based on `pie` and has no axes or orientation.

src/components/containers/SubplotAccordion.js
nicolaskruchten: the legend is merged, not overlapping, just like with a multi-pie scenario.

src/lib/customTraceType.js
nicolaskruchten: yeah I don't think it makes sense personally.

PR #936, opened 2019-06-13
column names to meta

nicolaskruchten:


nicolaskruchten:
This PR covers the immediate assignment, but not the pathways whereby the dataSourceOptions change under the hood.

nicolaskruchten:
Still need to handle the trace fold name issue

nicolaskruchten:
@VeraZab ready for review!

src/components/fields/DataSelector.js
```diff
@@ -94,6 +95,14 @@ export class UnconnectedDataSelector extends Component {
       fromSrc: this.context.srcConverters ? this.context.srcConverters.fromSrc : null,
     });
 
+    if (this.props.container.type) {
```
VeraZab: well, actually, it's possible not to provide trace.type in data, then plotly.js defaults to scatter..
maybe we should encode this in the editor.. this will only be potentially a problem for figures created with an api, where the person did not specify trace.type

VeraZab: oh actually nevermind, I think we'd be handling that here: https://github.com/plotly/react-chart-editor/blob/master/src/lib/customTraceType.js#L9

VeraZab:
looks good to me 💃 !

PR #946, opened 2019-07-03
add check array

oprstchn:
# bug fix
- If you try to split the data, an error occurs. It is PR of the correction for it.

Screenshot from 2019-07-03 12-36-25


nicolaskruchten:
Thanks for this! in order for our CI to run correctly I've migrated that commit into #947 and I'll merge there once it's green :)

nicolaskruchten:
This has been pushed up to NPM as 0.38.1 ... thanks again!

PR #950, opened 2019-07-15
fix: App> trouble with building on windows 10

cybermerlin:
fix: App> trouble with building on windows 10


nicolaskruchten:
Ah so here the idea is that you need to copy the `tpl` file to a real file: https://github.com/plotly/react-chart-editor#development-setup

PR #951, opened 2019-07-17
fix: lib.getFullTrace> fullData mb empty

cybermerlin:
I've exception on delete TraceFold because 1st remove data n then remove plot


nicolaskruchten:
Sorry I didn't see this PR until just now! thanks very much :)

PR #956, opened 2019-09-03
Disable / Enable attrs based on custom config

VeraZab:
I don't think I want to quite include the full whitelisting functionality in this pr. It can be a foundation for it, but more thought would have to be put into it, and it's not really the core of the functionality that we need atm.

Just to show an example of a case, say you enable very few attributes (just 'colorway' for example). Then you get situations like this:
Screen Shot 2019-09-05 at 4 53 56 PM

Where you have a panel that can add annotations, but nothing can be edited. It seems odd, feels like that menu item should not even be there. Thoughts @nicolaskruchten ?


src/lib/unpackPlotProps.js
```diff
@@ -29,9 +29,47 @@ export default function unpackPlotProps(props, context) {
   }
 
   let isVisible = false;
+  // Base visibility rules
   if (props.show || (fullValue !== void 0 && fullValue !== null)) {
     isVisible = true;
   }
+  // We can override base visibility rules with a customConfig
+  if (
```
nicolaskruchten: I would extract this out to a function somewhere so it's super-clear what the inputs are to the function, and it can be more easily unit-tested.

src/lib/__tests__/unpackPlotProps-test.js
```diff
@@ -0,0 +1,107 @@
+import {applyCustomConfigVisibility} from '../index';
```
VeraZab: @nicolaskruchten it would help me if you could think of more example cases to test here..

VeraZab:
this is the commit i'd like to double check my logic on: https://github.com/plotly/react-chart-editor/pull/956/commits/6f8473f72f18a06220a033e78ee22a4145a6ae65

I think what I've done is the right thing to do, but maybe we could discuss this tomorrow

src/components/fields/derived.js
```diff
@@ -171,23 +182,31 @@ export const ShowInLegend = connectToContainer(UnconnectedVisibilitySelect, {
 
 export const HistogramInfoVertical = connectToContainer(Info, {
   modifyPlotProps: (props, context, plotProps) => {
-    plotProps.isVisible =
-      context.fullContainer.type === 'histogram' && context.fullContainer.orientation === 'v';
+    plotProps.isVisible = isValidCustomConfigObject(context.customConfig)
+      ? plotProps.isVisible &&
+        context.fullContainer.type === 'histogram' &&
```
nicolaskruchten: extract `baseRule` ?

nicolaskruchten: throughout the file :)

src/lib/unpackPlotProps.js
```diff
@@ -1,9 +1,102 @@
+/* eslint-disable no-console */
+
 import nestedProperty from 'plotly.js/src/lib/nested_property';
 import isNumeric from 'fast-isnumeric';
 import {MULTI_VALUED, MULTI_VALUED_PLACEHOLDER} from './constants';
 
+const hasFullValue = fullValue => fullValue !== void 0 && fullValue !== null;
+
+export function isValidCustomConfigObject(customConfig) {
```
nicolaskruchten: consider renaming to something like `hasCustomVisibilityRules` ?

src/lib/unpackPlotProps.js
```diff
@@ -1,9 +1,102 @@
...
+      );
+      return false;
+    }
+
+    return true;
+  }
+  return false;
+}
+
+export function applyCustomConfigVisibility(props, fullValue, customConfig) {
```
nicolaskruchten: consider renaming to `isVisibleAccordingToCustomVisibilityRules` or soemthing shorter

nicolaskruchten: I would remove the `hasFullValue` checks from inside this function, so this is a "permitted but not enforced" function

src/lib/unpackPlotProps.js
```diff
@@ -28,10 +121,9 @@ export default function unpackPlotProps(props, context) {
     multiValued = true;
   }
 
-  let isVisible = false;
-  if (props.show || (fullValue !== void 0 && fullValue !== null)) {
-    isVisible = true;
-  }
+  const isVisible = isValidCustomConfigObject(customConfig)
```
nicolaskruchten: here I would use a rule like `isVisible = permittedByCustomRules() && (hasFullValue() || props.show)` (assuming that you remove the `hasFullValue` checked from inside the `pemitted` function

src/lib/__tests__/unpackPlotProps-test.js
```diff
@@ -0,0 +1,110 @@
...
+    };
+
+    const expectedResults = [false, true, false, true, false, false];
+
+    [
+      {attr: 'bg_color'},
+      {attr: 'colorscale'},
+      {attr: 'font.color'},
+      {attr: 'somethingElse'},
+      {attr: 'colorscale.title.font.color'},
```
nicolaskruchten: add `colorscale.ssomethingelse` ?

src/lib/__tests__/unpackPlotProps-test.js
```diff
@@ -0,0 +1,110 @@
...
+        blacklist: [{type: 'attrName', regex_match: 'color'}],
+        whitelist: [{type: 'attrName', regex_match: 'colorscale(?!.title.font.color|.tickcolor)'}],
+        order: ['blacklist', 'whitelist'],
+      },
+    };
+
+    const expectedResults = [false, true, false, true, false, false];
+
+    [
+      {attr: 'bg_color'},
```
nicolaskruchten: consider `{ 'bg_color': false }`

src/lib/__tests__/unpackPlotProps-test.js
```diff
@@ -0,0 +1,110 @@
+import {applyCustomConfigVisibility} from '../index';
+
+describe('applyCustomConfigVisibility', () => {
+  it('correctly blacklists attributes taking into account exceptions', () => {
+    const customConfig = {
+      visibility_rules: {
+        blacklist: [{type: 'attrName', regex_match: 'color'}],
+        whitelist: [{type: 'attrName', regex_match: 'colorscale(?!.title.font.color|.tickcolor)'}],
+        order: ['blacklist', 'whitelist'],
```
nicolaskruchten: show the behaviour for each test of each order of the lists

nicolaskruchten:
So when excluding color specifically we want to be able to say "yes to column mapping but no to specific color choosing" e.g. 

image

nicolaskruchten:
Similarly we want to be able to say "constant vs variable" but just hide the color-picker widget:

image

nicolaskruchten:
maybe we need to do widget-type-matching. like color-picker and font-family.

nicolaskruchten:
what happened to the image dropzone thing?

src/lib/unpackPlotProps.js
VeraZab: disregarding the props.show is what was making the dropzone dissapear, the logic above restores it

src/components/fields/derived.js
```diff
@@ -80,7 +87,9 @@ export const RangesliderVisible = connectToContainer(UnconnectedRadio, {
     if (!plotProps.fullValue) {
       plotProps.fullValue = false;
       plotProps.visible = false;
-      plotProps.isVisible = true;
+      plotProps.isVisible = hasValidCustomConfigVisibilityRules(context.customConfig)
+        ? computeCustomConfigVisibility(props, plotProps.fullValue, context.customConfig)
```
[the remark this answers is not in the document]
VeraZab: because in this case we're changing the fullValue, and that is taken into account in the computation

src/lib/connectToContainer.js
```diff
@@ -44,7 +45,7 @@ export default function connectToContainer(WrappedComponent, config = {}) {
     }
 
     setLocals(props, context) {
-      this.plotProps = unpackPlotProps(props, context);
+      this.plotProps = unpackPlotProps(props, context, WrappedComponent);
```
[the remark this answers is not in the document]
VeraZab: to get the component.displayName ... because now we're filtering based on that too..

src/lib/unpackPlotProps.js
```diff
@@ -1,9 +1,104 @@
...
+      console.error(errorMessage);
+      return false;
+    }
+
+    return true;
+  }
+  return false;
+}
+
+export function computeCustomConfigVisibility(props, fullValue, customConfig, wrappedComponent) {
```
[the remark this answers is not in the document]
VeraZab: but the outcome of this function can be true or false for visibility

src/lib/unpackPlotProps.js
```diff
@@ -1,9 +1,104 @@
...
+
 import nestedProperty from 'plotly.js/src/lib/nested_property';
 import isNumeric from 'fast-isnumeric';
 import {MULTI_VALUED, MULTI_VALUED_PLACEHOLDER} from './constants';
 
-export default function unpackPlotProps(props, context) {
-  const {container, getValObject, defaultContainer, updateContainer} = context;
+const hasFullValue = fullValue => fullValue !== void 0 && fullValue !== null;
+
+export function hasValidCustomConfigVisibilityRules(customConfig) {
```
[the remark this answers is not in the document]
VeraZab: but it actually checks the validity too

src/lib/unpackPlotProps.js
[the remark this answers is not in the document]
nicolaskruchten: the "positive" version of this is `permittedByCustomVisibilityRules`

src/lib/unpackPlotProps.js
```diff
@@ -1,7 +1,109 @@
...
+  }
+
+  if (customConfig.visibility_rules.whitelist) {
+    isVisible = customConfig.visibility_rules.whitelist.some(passesTest);
+  }
+
+  return isVisible;
+}
+
+export function validateVisibilityWithCustomConfig(
```
nicolaskruchten: rename to boolean-like name... `isVisibleGivenCustomConfig` or something like this...?

VeraZab:
I think I addressed all the comments, please let me know if I can merge and release @nicolaskruchten.

src/components/containers/__tests__/ConnectedContainersVisibility-test.js
```diff
@@ -74,6 +74,56 @@ describe('Basic PlotlySection rules', () => {
...
+            <PlotlySection attr="title">
+              <Numeric attr="title.font.color" />
+            </PlotlySection>
+          </LayoutPanel>
+        </TestEditor>
+      );
+
+      it('HIDES Field based on customConfig', () =>
+        expect(wrapper.find('input').length).toEqual(0));
+      it('SHOWS PlotlySection if it has an attr that is accepted by customConfig', () =>
```
nicolaskruchten: shouldn't this section be hidden because it's empty?

VeraZab: I could make it so, if i remove a part of the current logic that we have: https://github.com/plotly/react-chart-editor/blob/master/src/components/containers/PlotlySection.js#L51
(the || , just make it return if there's no child)

It didn't seem necessary for the current config we're trying to get right, and it's previously existing logic, that is not entirely false. (I mean it appears because we did put an attr, that is accepted by customConfig on a PlotlySection)

nicolaskruchten: to make this test easier to understand, can we make it such that there is another widget in the section with the same attr as the section?

nicolaskruchten:
💃

PR #961, opened 2019-09-16
Full Mapbox support

nicolaskruchten:


VeraZab:
in the dev app, could we remove the mapbox access token please? i often end up losing that accessToken file

VeraZab:
testing on this branch: https://github.com/plotly/react-chart-editor/pull/962
but since you're probably going to use this one to merge and make adjustments, commenting here.

I feel like subplots are not working anymore, unless im missing something
Screen Shot 2019-09-17 at 11 57 35 AM
Screen Shot 2019-09-17 at 11 57 24 AM

VeraZab:
almost all of the marker shapes are not working for me, and even the working ones don't show up in the right color
Screen Shot 2019-09-17 at 1 05 44 PM

nicolaskruchten:
Well as of this PR you can set the token to an empty string :)

nicolaskruchten:
Meaning you can just do `cp accessTokens.js.tpl accessTokens.js` and it'll "just work" without the Mapbox-service features

VeraZab:
As discussed on slack: 
- for the scattermapbox trace, looks like changing the base map tile source, when layers have already been added, does not keep the layers. Even deleting and adding back the layers does not work.
- also, i seem to have gotten a few times into a state when deleting my trace (Ex choropleth tile map) and adding another one (ex. density tile map). And then adding lat/lon/z, did not do anything to the map. no trace was drawn. sometimes i get into that state, sometimes i don't..

dev/App.js
```diff
@@ -14,7 +14,7 @@ import 'brace/theme/textmate';
 // https://github.com/plotly/react-chart-editor#mapbox-access-tokens
 import ACCESS_TOKENS from '../accessTokens';
 
-import {customConfigTest} from '../src/__stories__';
+// import {customConfigTest} from '../src/__stories__';
```
VeraZab: I put it in `../src/__stories`, because when i had it in `dev/App.js` the imports that file has were causing issues with the percy tests. I guess this could go into it's separate file. But yes we can just comment out. Not exactly sure though what's causing the issue.. maybe that we reimport plotly.js in that storybook file, from a different place..

VeraZab:
code looks good to me otherwise 💃

VeraZab:
squash commits before merging?

nicolaskruchten:
yes, for sure :)

PR #962, opened 2019-09-17
Mapboxwip

nicolaskruchten:


VeraZab:
i'll comment on what i see here as this is the branch i'm testing in my dev env
so: 

almost all of the marker shapes are not working for me, and even the working ones don't show up in the right color
Screen Shot 2019-09-17 at 1 05 44 PM

nicolaskruchten:
Yeah, marker symbols in Mapbox were broken before this PR. I could try to fix it.

Issue #1019, opened 2020-03-09
npm start error of development setup


chl87256:
![error](https://user-images.githubusercontent.com/61171990/76187822-bcd4f980-6219-11ea-83cd-c33045ef148a.PNG)
I did a development setup of react-chart-editor.
The installation proceeded to npm install, but after npm start, the error shown in the picture was displayed.
I tried several things but couldn't find the answer.
I thought it was a babel problem, so I upgraded and modified the index.js file, but it didn't work.
If you know how, please let me know. So thank you so much.

dmt0:
Would you mind trying `yarn install` instead of `npm install`?
What is your node.js version?

chl87256:
> ould you mind trying `yarn install` instead of

I'm using centos7 so I can't use yarn install. The version of node.js is 6.17.1

dmt0:
node.js version is likely a problem. A lot of the dependencies are no longer supporting anything below v10. We're doing development with v12.

chl87256:
> node.js version is likely a problem. A lot of the dependencies are no longer supporting anything below v10. We're doing development with v12.

thank you!! Yeah! I upgrade node js and fix other errors so it works fine.

PR #1020, opened 2020-03-10
Require Node.js v10 or above

dmt0:
Avoid confusion by requiring node v10 or above.
Closes https://github.com/plotly/react-chart-editor/issues/1019

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,9 @@
     "react": ">15",
     "react-dom": ">15"
   },
+  "engines": {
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
Does this impact Chart Studio?

dmt0:
We upgraded to v12 across the board. 8 is officially unsupported since beginning of year. 10 is the next LTS. So there you go.

nicolaskruchten:
OK 💃

PR #1083, opened 2020-09-30
Packages + plotly.js + version bump

dmt0:
- Package updates
- Plotly.js 1.56 support
- Version bump


dmt0:
@nicolaskruchten Can you take a look at Percy, is everything the way you expect it to be?

dmt0:
(got 💃 over slack)

Issue #1168, opened 2022-08-29
missing translation keys


WillyKirchner:
Hello, 
While using the chart editor, i found out that these two expressions can't be translated. This could be fixed by changing them into strings.

https://github.com/plotly/react-chart-editor/blob/c0586bb43f381bb6794050394982c76728fff050/src/components/fields/ColorPicker.js#L20
https://github.com/plotly/react-chart-editor/blob/c0586bb43f381bb6794050394982c76728fff050/src/components/fields/DataSelector.js#L132

Issue #1009, opened 2020-02-05
sass-loader past 7.1.0 does not work


dmt0:
Upgrading sass-loader past 7.1.0 breaks build silently.

Issue #854, opened 2019-02-26
Dev app doesn't work in IE


dmt0:
`fetch` needs to be polyfilled

VeraZab:
and i think the css is not working there too, we're compiling 2 versions of sass, one for IE specifically: https://github.com/plotly/react-chart-editor/blob/master/package.json#L114

dmt0:
Yes, it's all those custom properties like `var(--sidebar-width)` - they're not supported in IE, our tools used to take care of those though.

dmt0:
Were we ever able to test RCE in IE without embedding it into another project?

VeraZab:
I always was testing from streambed..

nicolaskruchten:
Did you see this part of the readme? https://github.com/plotly/react-chart-editor#styling-the-defaulteditor--and-the-built-in-components

If the dev/demo apps could work in IE that would be helpful IMO

chl8725:
> Did you see this part of the readme? https://github.com/plotly/react-chart-editor#styling-the-defaulteditor--and-the-built-in-components
> 
> If the dev/demo apps could work in IE that would be helpful IMO

I can't display in ie.
But go to the link you said
I found import'react-chart-editor/lib/react-chart-editor.ie.min.css' but there is no react-chart-editor.ie.min.css file
where is the file?

PR #1172, opened 2023-09-07
Refresh

dmt0:
- Updates to dev tools - SASS, Percy/Storybook, others
- Fix CI
- Drop IE support
- Use Volta
- Allow Plotly.js V2 as peer dep
- Fix minor bugs
- Minify production build
- Update demo app (used in github pages)
- Closes https://github.com/plotly/react-chart-editor/issues/1168
- Closes https://github.com/plotly/react-chart-editor/issues/1009
- Closes https://github.com/plotly/react-chart-editor/issues/854


dmt0:
Hey @alexcjohnson, this one is ready for review. Once this is reviewed and merged I'll make a release.

Looks like Percy didn't keep the old screenshots and counts these as brand new initial ones. Would be great if you'd approve them, so we have a reference for future work.

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ jobs:
   build:
     docker:
       # specify the version you desire here
-      - image: circleci/node:12.13.1-browsers
+      - image: cimg/node:14.19-browsers
```
alexcjohnson, line 6: Thanks for updating to `cimg`. But node 14 is still pretty old, most of our projects have moved on to at least 16, some to 18 (and now I see 20 is out even).

babel.config.json
```diff
@@ -0,0 +1,32 @@
...
+        "alias": {
+          "components": "./src/components",
+          "lib": "./src/lib",
+          "styles": "./src/styles"
+        }
+      }
+    ]
+  ]
+  // "comments": false,
+  // "minified": true,
```
alexcjohnson, line 31:
```suggestion
```
or uncomment, whichever you actually intended.

src/styles/_movement.scss
```diff
@@ -41,7 +41,8 @@
   }
 }
 
-.animate {
+.animate,
+%animate {
```
alexcjohnson, line 45: where is `%animate` used? (and thanks for teaching me about [placeholder selectors](https://sass-lang.com/documentation/style-rules/placeholder-selectors/), new concept to me! 🧑‍🏫 )

package.json
```diff
@@ -135,6 +139,15 @@
     "url": "https://github.com/plotly/react-chart-editor.git"
   },
   "browserslist": [
-    "ie 11"
-  ]
+    "defaults",
+    "not IE 11"
```
alexcjohnson, line 143:
```suggestion
    "last 8 years and not dead"
```
I'm not quite sure what `"defaults"` means, but as I recall it was a little more aggressive about dropping older browsers than we were comfortable with in some contexts. The above is what we switched to in eg [dash-renderer](https://github.com/plotly/dash/blob/6eaf2e17c25f7ca1847c41aafeb18e87c586cb9f/dash/dash-renderer/package.json#L90) when we dropped IE support.

alexcjohnson:
@dmt0 this looks great. Just had a couple of small suggestions, and it looks like whatever the issue was with Percy has been resolved and it's happy now ☺️

src/styles/_movement.scss
dmt0: You have the original developer of the feature to thank :)
We currently only use it to fade out modals on close, but there have been plans to animate more parts of the UI.

.circleci/config.yml
dmt0: I totally agree that 14 is old. IIRC this was an easy upgrade and upgrading to anything newer turned out to be non-trivial. I figured it doesn't matter that much since it's only used for CI. I can sure take a look at it though.

dmt0: Old Storybook doesn't want to work with newer Node. I've started the upgrade process. It's indeed non-trivial.

alexcjohnson: ah ok - would be nice to update for sure but if this is too onerous I won't insist.

src/styles/_movement.scss
alexcjohnson: I was just expecting to see an `@extend %animate` along with this, but I don't, am I misunderstanding how `%animate` would be used?

.circleci/config.yml
dmt0: Yeah, I'd argue it's best to delay this upgrade until https://github.com/storybookjs/storybook/issues/9828 is resolved. That's supposed to happen with Storybook V8 release.

As of now I've sunk a couple of hours into getting Storybook V7 to work. The result is pretty ugly (you can see it in my last commit). And now that Storybook works, Percy refuses to work, so I opened a support ticket with Percy to sort it out.

Node version in CI doesn't affect us much, as I haven't yet bumped into any packages that dropped support for it (except for Storybook).

I can take that Storybook commit and cherry-pick it into a separate PR in order not to loose that work.

src/styles/_movement.scss
dmt0: You're right! Animation is actually not there. I was using automated tools to do most of the SASS upgrade and haven't noticed how the animations disappeared. Fix incoming.

.circleci/config.yml
dmt0: Percy fixed by downgrading it. So we have a build. I'll let you decide if you wanna keep this Node + Storybook upgrade.

alexcjohnson: Thanks for fighting through that - let's keep to the latest Node + Storybook, seems like that's the bigger change than having to push Percy back a minor and will keep us closer to the latest versions overall.

alexcjohnson (approved):
💃 great stuff @dmt0!

